PR #24, opened 2022-09-23
Dash integration

pmitros:
@bradley-erickson This PR successfully serves your dashboards from dash (albeit without things like CSS). It also fixes several bugs in the main codebase (which, unfortunately, other pieces of the codebase relied on). It's not ready for merge, but it is at a stage where it's ready for a general review of the direction it's going.


bradley-erickson:
@pmitros what are the steps to get this to serve the dashboard? Do I just need to spin up WO and navigate to /dashboard?
Is there a specific piece of the code I should be working on?

pmitros:
@bradley-erickson I'm not trying to hand off work, so much as to ask for a general code review and to sanity-check the direction I'm taking. 

`/admin/status` is generally a good place to start. Browsers like Firefox render `json` nicely and interactively. Under `modules/dash_pages`, you'll see the installed `dash` modules. LO comes with a dummy test case under `learning_observer/dash/test`, while your dashboard (if you run `python setup.py develop`) is under `http://0.0.0.0:8892/wo_dash/dash/dashboard`

I think the code will be more mature to collaborate upon with a few hours of work, whenever I squeeze that in. I had a bug where the URLs were using slightly the wrong name from things, since I was extracting the wrong package name somewhere. I think I need to fix that up throughout the system before merging your changes and my changes asynchronously will be easy.

pmitros:
@bradley-erickson Okay. I stepped through the URL issues and pushed a minor commit. I think this would be a good time to start to work together and discuss next steps.

PR #27, opened 2022-12-20
Merge System Fixes

DrLynch:
This PR brings in the system fixes including renaming the cases and fixing the roster alignment for the writing data.


pmitros:
Thank you for this pull request. Would it be possible to restructure this as several pull requests? Atomic PRs might be:

1. Change from camel case to underscores. This is a high priority since we'll get code drift if we don't merge this.
2. Small PRs for narrow changes, such as doc fixes in the Chrome extension or documentation of Kafka, adding the dashboard to `module.py`, etc. These are second priority, since these would be nice to get in.
3. Dev-ops changes. Those should probably go under `/devops/`, and should not include things hardcoded to NCSU. It makes sense to add RHEL to this repo, but the NCSU-specific work should probably be in its own repo. 

For devops, I'll mention that the `ansible` stuff in my repo is nicer than your stuff, since servers can be configured in one command and we can provision / de-provision a whole flock of them as needed for different tasks (which is rather faster and more maintainable than with a `README` file describing steps). It also allows us to keep track of what's deployed when, which is well-aligned with the open science goals of the project. 

This is probably the lowest-priority since it's a lot of work, and your current version is pretty orthogonal to what's there.

It would be helpful to remove noise changes, such as adding/removal of `print` statements, adding / removing random lines (unless for pycodestyle/pep8/pylint alignment), temporary debugging changes to `kvs.py`, commented-out lines, the `testcode` directory, etc. This stuff should be `git stash`ed but remain on your local system.

DrLynch:
Piotr, I will see if I can find the time to split things into separate PRs.  

I have already excised the excess printing code. The commented lines remain are assisting in debugging and getting things working is my first concern.  At this point given the level of hacking in the system some commented cruft is not the major problem.  Both Rosters and KVS contain hacking and will likely change more.     Ansible, while it can be useful, is simply not a goal right now. It would be a lot of work for little gain.  We are not doing parallel provisioning of servers, but we are making architecture changes to accommodate the NLP code and the draft dashboards.  That to me is much higher priority.

pmitros:
@DrLynch 

* Following `pycodestyle` and not committing spurious changes saves more time than it costs. A messy commit record makes it easy to accidentally introduce bugs, and hard to understand what happened. 
* I would not commit NCSU-specific dev ops here. It's fine that this is hacked together, but there's no reason to have it in this repo.
* Using `ansible` is not a priority right now, but proper dev ops saves time in the long term. I do maintain a fleet of servers with `ansible` already. The infrastructure I have there saves time in the long term, especially when debugging or trying to understand what happened. There's no reason to commit a hackish infrastructure when a more robust one is already in the repo. Two infrastructures isn't better than one.

DrLynch:
Piotr, I agree on committing spurious code. That is why I avoid doing it in general. Commented code for debugging however is already all over the system so if we plan to establish a standard perhaps we should start with a cleanup.  Ansible will have to wait until I have the time to change to it or a student capable of doing so.  At the moment however the fixes take higher priority to me.  In any event I will close this and deal with the changes as I have time.

pmitros:
@DrLynch That's not actually true. Commented code for debugging is very much **not** all over the system. I work hard to `stash` debug code before I commit. Commented code would be okay at a much earlier stage for a project.

We were mostly compliant with `pycodestyle`, and I pushed a commit to fix most of the remaining gaps. You can run `make codestyle` if you would like to check where we are at any point. We have one issue with use of `lambda`s, where I disagree with `pycodestyle` but I'll clean that up as well at some point. Occasionally a few issues drift in, but I do period clean-ups like this one. 

The basic point is that it's important to develop in such a way that the code gets better. I'll try to add more about code quality in a README file at some point, and perhaps that can serve as a basis of discussion to see how that compares to how you'd like to work. 

I would like to get the changes are camel case merged, though, so we don't have code drift. That's the main place where keeping changes outside of mainline carries high costs down-the-line.

PR #30, opened 2022-12-25
Pluginfix

DrLynch:
Fixing issue with undefined state, missing bcrypt, and documenting a potential error.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ aiohttp_session
 aiohttp_wsgi
 asyncio_redis
 asyncpg
+bcrypt
 cryptography
 cython
 dash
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -6,6 +6,7 @@ aiohttp_session
 aiohttp_wsgi
 asyncio_redis
 asyncpg
+bcrypt
```
pmitros, line 9: We include `py-bcrypt` below. 

RANT:  I have not seen **any** set of requirements which works **everywhere**. There are small API differences -- especially around which things are byte strings versus strings, and whatever I've done breaks somewhere and is followed by a series of `pip install` commands. We should either switch to a more compatible library, have special-case code detecting which version of `bcrypt` we have, or take time to figure this out better otherwise.

bradley-erickson:
This pull request is now stale. Multiple pull requests with the same code were opened up. This results in future pull requests showing those items as changes since they are made with a different version of Master. This is messy to deal with. The avoid this in the future **specify different features to different pull requests**.

PR #32, opened 2023-01-27
Dami top right btn

bradley-erickson:


learning_observer/prototypes/dash_wo_teacher_dashboard/wo_dash/assets/styles.css
```diff
@@ -71,7 +71,11 @@
     box-shadow: 0px 0px 5px var(--bs-gray-300);
     border-radius: 0.4rem;
 }
-
+.btn-menu-outline-dark button {
```
pmitros, line 74: FYI: CSS is better with comments explaining what these things do. Not a showstopper for merge.

PR #36, opened 2023-02-03
Full nlp options

bradley-erickson:
- Implemented extra views capability
- Added indicator view to writing observer to see the options
- Client now reads options from WO then builds settings panel
- Implemented a lot more indicators to WO

### DIFF
--- a/learning_observer/learning_observer/module_loader.py
+++ b/learning_observer/learning_observer/module_loader.py
@@ -35,6 +35,7 @@ THIRD_PARTY = {}
 STATIC_REPOS = {}
 STUDENT_DASHBOARDS = []
 COURSE_DASHBOARDS = []
+EXTRA_VIEWS = []
 
 # Additional calls, primarily for metadata
 AJAX = {}
@@ -49,10 +50,9 @@ def extra_views():
     will use the same API, provide backwards-compatibility, but also
     act as a place for things which aren't dashboards. Modules ought
     to be able to define random views.
-
-    To Be Implemented
     '''
-    return []
+    load_modules()
+    return EXTRA_VIEWS
 
 
 def student_dashboards():
@@ -325,6 +325,19 @@ def load_dashboards(component_name, module):
         debug_log(f"Component {component_name} has no dashboards")
 
 
+def load_extra_views(component_name, module):
+    '''
+    '''
+    extras = False
+    if hasattr(module, 'EXTRA_VIEWS'):
+        debug_log(f'Loading extra views from {component_name}')
+        EXTRA_VIEWS.extend(module.EXTRA_VIEWS)
+        extras = True
+
+    if not extras:
+        debug_log(f'Component{component_name} has no extra views')
+
+
 def register_3rd_party(component_name, module):
     '''
     Register 3rd party components the module needs.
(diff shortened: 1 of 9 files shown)


learning_observer/learning_observer/routes.py
```diff
@@ -128,6 +128,15 @@ def tracemalloc_handler(request):
     #             "/views/" + extra_views[view]['url'],
     #             handler=extra_views[view]['function'])
     #     ])
+    extra_views = learning_observer.module_loader.extra_views()
+    for view in extra_views:
+        app.add_routes([
+            aiohttp.web.get(
+                f'/views/{view.get("url")}',
```
pmitros, line 135: Should be namespaced: `{view["module"]}/{view["suburl"]}`

Also, `.get` is only helpful if there's a default or similar.

learning_observer/learning_observer/routes.py
```diff
@@ -128,6 +128,15 @@ def tracemalloc_handler(request):
...
+                lambda x: aiohttp.web.json_response(view['data'])
```
pmitros, line 136: `data` is a bit generic. `view['handler']`?

pmitros: For debugging, I'd also prefer a decorator or at least a `def`, so we have a reasonable stack trace. I suspect we'll be debugging here a fair bit.

learning_observer/prototypes/dash_wo_teacher_dashboard/wo_dash/pages/assignment_page/settings_defaults.py
```diff
@@ -7,19 +7,26 @@
...
+            'statements_of_opinion', 'statements_of_fact',
+            'direct_speech_verbs', 'indirect_speech_quotation',
+            'in_past_tense', 'propositional_attitudes', 'social_awareness',
+            'transitions',
+        ],
+        'selected': ['main_idea_sentences']
     },
     'indicators': {
-        'options': ['is_academic', 'vwp_source', 'vwp_attribution', 'vwp_cite', 'vwp_interactive', 'vwp_argumentword', 'vwp_evaluation'],
+        'options': ['academic_language', 'informal_language', 'argument_words', 'latinate_words', 'information_sources', 'attributions', 'citations'],
```
pmitros, line 29: This is much nicer. Thank you.

modules/writing_observer/writing_observer/awe_nlp.py
```diff
@@ -102,28 +102,21 @@ def process_text(text, options=[]):
     doc = nlp(text)
```
pmitros, line 102: I'm very confused here. This seems to add some, but not all, of the code from my commits. There's an `options=[]` just above. Does this need to be rebased? Or what's going on here?

learning_observer/learning_observer/routes.py
pmitros: Oh. I just read below. This actually is just returning a JSON data object. That's unusual, but kind of makes sense. I still think we need something better than `data` here. Perhaps `static_json` or similar?

learning_observer/learning_observer/module_loader.py
```diff
@@ -325,6 +325,19 @@ def load_dashboards(component_name, module):
         debug_log(f"Component {component_name} has no dashboards")
 
 
+def load_extra_views(component_name, module):
+    '''
+    '''
+    extras = False
+    if hasattr(module, 'EXTRA_VIEWS'):
+        debug_log(f'Loading extra views from {component_name}')
+        EXTRA_VIEWS.extend(module.EXTRA_VIEWS)
```
pmitros, line 334: We want to make sure the we include the `component_name` on load here.

PR #37, opened 2023-02-10
Added nlp cache

bradley-erickson:
Now we pass writing_data down to process_texts to properly cache the options.

### DIFF
--- a/modules/writing_observer/writing_observer/awe_nlp.py
+++ b/modules/writing_observer/writing_observer/awe_nlp.py
@@ -25,7 +25,8 @@ import time
 import warnings
 
 import writing_observer.nlp_indicators
-from learning_observer import kvs, util
+import learning_observer.kvs
+import learning_observer.util
 
 RUN_MODES = enum.Enum('RUN_MODES', 'MULTIPROCESSING SERIAL')
 
@@ -217,13 +218,13 @@ async def process_texts(writing_data, options=None, mode=RUN_MODES.MULTIPROCESSI
     results = []
     need_processing = []
     needed_options = set()
-    cache = kvs.KVS()
+    cache = learning_observer.kvs.KVS()
 
     for writing in writing_data:
         text = writing.get('text', '')
         if len(text) == 0:
             continue
-        text_hash = 'NLP_CACHE_' + util.secure_hash(text.encode('utf-8'))
+        text_hash = 'NLP_CACHE_' + learning_observer.util.secure_hash(text.encode('utf-8'))
         text_cache_data = await cache[text_hash]
         if text_cache_data is None:
             text_cache_data = {}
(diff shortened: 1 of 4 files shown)


modules/writing_observer/writing_observer/awe_nlp.py
```diff
@@ -190,6 +197,68 @@ async def process_texts_parallel(texts, options=None):
...
+    '''
+    Create cache key based on text
+    '''
+    hash_key = hashlib.sha1(text.encode('utf-8')).hexdigest()
+    return 'NLP_CACHE_' + hash_key
+
+
+async def process_texts(writing_data, options=None, mode=RUN_MODES.MULTIPROCESSING):
+    '''
+    Process texts with cachinng
```
pmitros, line 211: typo

modules/writing_observer/writing_observer/awe_nlp.py
```diff
@@ -190,6 +197,68 @@ async def process_texts_parallel(texts, options=None):
 
     return annotated
 
+
+def hash_cache_key(text):
```
pmitros, line 201: I don't think we want to define yet another hash function. I would

* Take the `secure_hash` and `insecure_hash` functions in `log_event`
* Move them into `util.py`
* Use the secure one here

If there is an issue with hash security, we do want to be able to address it globally, system-wide.

modules/writing_observer/writing_observer/awe_nlp.py
```diff
@@ -190,6 +197,68 @@ async def process_texts_parallel(texts, options=None):
...
+        RUN_MODES.SERIAL: process_texts_serial
+    }
+
+    results = []
+    need_processing = []
+    needed_options = set()
+    cache = kvs.KVS()
+
+    for writing in writing_data:
+        text = writing.get('text', '')
```
pmitros, line 231: I'd consider treating the empty string case as a special case.

modules/writing_observer/writing_observer/awe_nlp.py
```diff
@@ -190,6 +197,68 @@ async def process_texts_parallel(texts, options=None):
...
+        
+        for annotated_text, single_doc in zip(annotated_texts, need_processing):
+            if annotated_text != "Error":
+                single_doc.update(annotated_text)
+                text_hash = hash_cache_key(single_doc.get('text', ''))
+                # which is the lesser of two evils?
+                # we store the student key in the cache
+                # OR 
+                # we remove it before storing (uses dict comprehension)
+                new_cache = {k: v for k, v in single_doc.items() if k != 'student'}
```
pmitros, line 256: This is probably not the right format. We would like to be able to add more metadata, such as:
- When this was computed. 
- If a computation is running

PR #38, opened 2023-02-14
Stub code work

bradley-erickson:
- Adjust stub code to use start index/offset (instead of start/end index)
- Created a process texts method for using the stub code to generate results
- Added entrypoint to determine if awe_components are installed and if the use_nlp setting is true

### DIFF
--- a/modules/writing_observer/writing_observer/aggregator.py
+++ b/modules/writing_observer/writing_observer/aggregator.py
@@ -1,5 +1,9 @@
+import importlib.util
 import time
+
+import learning_observer.settings
 import learning_observer.util
+
 
 def excerpt_active_text(
     text, cursor_position,
@@ -178,7 +182,14 @@ async def merge_with_student_data(writing_data, student_data):
     return writing_data
 
 
-import writing_observer.awe_nlp
+if importlib.util.find_spec('awe_components') is not None and\
+    learning_observer.settings.module_setting('writing_observer', 'use_nlp', False):
+    import writing_observer.awe_nlp
+    processor = writing_observer.awe_nlp.process_texts
+else:
+    import writing_observer.stub_nlp
+    processor = writing_observer.stub_nlp.process_texts
+
 
 async def latest_data(student_data, options=None):
     '''
@@ -189,5 +200,5 @@ async def latest_data(student_data, options=None):
     writing_data = await get_latest_student_documents(student_data)
     writing_data = await remove_extra_data(writing_data)
     writing_data = await merge_with_student_data(writing_data, student_data)
-    writing_data = await writing_observer.awe_nlp.process_texts(writing_data, options)
+    writing_data = await processor(writing_data, options)
     return {'latest_writing_data': writing_data}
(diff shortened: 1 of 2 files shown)


modules/writing_observer/writing_observer/aggregator.py
```diff
@@ -178,7 +182,14 @@ async def merge_with_student_data(writing_data, student_data):
     return writing_data
 
 
-import writing_observer.awe_nlp
```
pmitros, line 181: What is the purpose of this line?

I'm not sure I'm reading this correctly, but if I am, this is very bad form. 

If we expect to find AWE Workbench and cannot find it, we want to fail early. A failure on startup allows us to fix the bug on a deployed server. 

A silent failure will mean we're delivering bad NLP to teachers. 

What I generally do is along the lines of:

`if [can't find module]: print([Error message] exit()`

The error message contains instructions for:

1. Installing the module
2. Enabling the stub from the config file

modules/writing_observer/writing_observer/stub_nlp.py
```diff
@@ -44,7 +46,7 @@ def select_random_segments(text, segments=3, segment_length=3, seed=0):
     word_boundaries = [match.start() for match in re.finditer(r"\b\w+\b", text)]
     word_boundary_count = len(word_boundaries)
     selected_indices = limited_sample(word_boundary_count - segment_length, segments)
-    segments_positions = [(word_boundaries[index], word_boundaries[index + segment_length]) for index in selected_indices]
```
pmitros, line 47: PEP8: Spaces around minus

modules/writing_observer/writing_observer/stub_nlp.py
```diff
@@ -97,7 +100,7 @@ def show_selections(text, segments):
     Returns:
         list: A list of strings, each representing a selected segment of the input text.
     '''
-    return [text[a:b] for a, b in segments]
```
pmitros, line 100: PEP8: Spaces around operator

modules/writing_observer/writing_observer/stub_nlp.py
```diff
@@ -117,3 +120,49 @@ def find_keywords_in_text(text, keywords):
...
+        if len(text) == 0:
+            continue
+        results = {}
+        for option in options:
+            if option not in writing_observer.nlp_indicators.INDICATORS:
+                continue
+            indicator = writing_observer.nlp_indicators.INDICATORS[option]
+            (id, label, infoType, select, filterInfo, summaryType) = indicator
+            results[id] = {}
+            random.seed(id)
```
pmitros, line 143: If you're doing a seed, you should save the RNG state and restore it. Otherwise, this will cause cascading Heisenbugs anywhere else we use `random`. 

Indeed, this can escalate security issues if, for example, some code uses `random` instead of `urandom`. Normally, that's a hard-to-exploit security issue. If an attacker knows the seed, it becomes an easy-to-exploit security issue.

pmitros:
Looks pretty good, modulo the issues indicated. In addition, the docstrings should be brought in-line with the new format.

Thank you for the contribution.

PR #41, opened 2023-02-18
Merging in creds.yaml tweak.

DrLynch:
This provides a simple tweak of the creds.yaml file to support deployment on chromebooks.


bradley-erickson:
@pmitros cleaned up some styling and added in the modules/writing observer to the example as well.

PR #55, opened 2023-03-27
Fetching data from Google locally

bradley-erickson:
Creating a draft pull request to review as we go with this.
- Grabs document text from google api instead of reconstructing text
- Confirmed it when passing a specific google doc id
- Added in requests to aggregator function
- Updated link to doc on student card to only show up when they have a latest doc

### DIFF
--- a/learning_observer/learning_observer/auth/social_sso.py
+++ b/learning_observer/learning_observer/auth/social_sso.py
@@ -90,7 +90,7 @@ async def _google(request):
         return {}
 
     hostname = settings.settings['hostname']
-    protocol = settings.settings.get('protocol', 'http')
+    protocol = settings.settings.get('protocol', 'https')
     common_params = {
         'client_id': settings.settings['auth']['google_oauth']['web']['client_id'],
         'redirect_uri': f"{protocol}://{hostname}/auth/login/google"
--- a/learning_observer/learning_observer/auth/social_sso.py
+++ b/learning_observer/learning_observer/auth/social_sso.py
@@ -89,12 +89,11 @@ async def _google(request):
     if 'error' in request.query:
         return {}
 
+    hostname = settings.settings['hostname']
+    protocol = settings.settings.get('protocol', 'http')
     common_params = {
         'client_id': settings.settings['auth']['google_oauth']['web']['client_id'],
-        'redirect_uri': "{protocol}://{hostname}/auth/login/google".format(
-            protocol=settings.settings['protocol'],
-            hostname=settings.settings['hostname']
-        )
+        'redirect_uri': f"{protocol}://{hostname}/auth/login/google"
     }
 
     # Step 1: redirect to get code
--- a/learning_observer/learning_observer/auth/social_sso.py
+++ b/learning_observer/learning_observer/auth/social_sso.py
@@ -91,7 +91,8 @@ async def _google(request):
 
     common_params = {
         'client_id': settings.settings['auth']['google_oauth']['web']['client_id'],
-        'redirect_uri': "https://{hostname}/auth/login/google".format(
+        'redirect_uri': "{protocol}://{hostname}/auth/login/google".format(
+            protocol=settings.settings['protocol'],
             hostname=settings.settings['hostname']
         )
     }
--- a/learning_observer/learning_observer/cache.py
+++ b/learning_observer/learning_observer/cache.py
@@ -1,0 +1,27 @@
+from functools import wraps
+import json
+
+import learning_observer.kvs
+
+
+def create_key_from_args(*args, **kwargs):
+    key_dict = {'args': args, 'kwargs': kwargs}
+    key_str = json.dumps(key_dict, sort_keys=True)
+    return key_str
+
+
+def async_memoization():
+    # TODO how should we default here if memoization_cache doesn't exist?
+    cache_backend = learning_observer.kvs.KVS_DICT['memoization_cache']()
+
+    def decorator(func):
+        @wraps(func)
+        async def wrapper(*args, **kwargs):
+            key = create_key_from_args(args, kwargs)
+            if key in await cache_backend.keys():
+                return await cache_backend[key]
+            result = await func(*args, **kwargs)
+            await cache_backend.set(key, result)
+            return result
+        return wrapper
+    return decorator
--- a/learning_observer/learning_observer/cache.py
+++ b/learning_observer/learning_observer/cache.py
@@ -1,4 +1,4 @@
-from functools import wraps
+import functools
 import json
 
 import learning_observer.kvs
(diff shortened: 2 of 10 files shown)


learning_observer/learning_observer/auth/social_sso.py
```diff
@@ -91,7 +91,8 @@ async def _google(request):
 
     common_params = {
         'client_id': settings.settings['auth']['google_oauth']['web']['client_id'],
-        'redirect_uri': "https://{hostname}/auth/login/google".format(
+        'redirect_uri': "{protocol}://{hostname}/auth/login/google".format(
+            protocol=settings.settings['protocol'],
             hostname=settings.settings['hostname']
```
pmitros, line 96: My preference would be `hostname = settings.settings['hostname']`, `protocol=settings.settings.get('protocol', "http")` and then a format string: 

`f"{protocol}://{hostname}/auth/login/google"`

learning_observer/learning_observer/dashboard.py
```diff
@@ -222,7 +222,7 @@ async def student_state_fetcher():
                     sa_module,
                     {sa_helpers.KeyField.STUDENT: student_id},
                     sa_helpers.KeyStateType.EXTERNAL)
-                debug_log(key)
```
pmitros, line 225: Should probably be removed, rather than commented.

learning_observer/learning_observer/rosters.py
```diff
@@ -412,6 +412,10 @@ async def courseroster(request, course_id):
     List all of the students in a course: Helper
     '''
     if settings.settings['roster_data']['source'] in ["google_api"]:
+        if 'error' in request:
+            raise learning_observer.prestartup.StartupCheck(
+                f'Error in request: {request}'
+            )
```
pmitros, line 418: This code is incorrect:

1. `StartupCheck` shouldn't be raised outside of startup.
2. I don't know what `'error' in request` means. This should be commented or explained.

pmitros:
I found one error and two nits in the initial version. 

I *really* don't like passing around the request object as you're doing it; the modules should be agnostic to the web framework being used. I don't know off-hand a better way to do this, though. It does break a lot of abstraction barriers.

Perhaps one alternative is to pass through a black box object which provides for context-specific callbacks. We should discuss.

bradley-erickson:
This code includes some clean up around the google apis and provides an easy way to access documents in writing observer.

learning_observer/learning_observer/auth/social_sso.py
```diff
@@ -89,11 +89,11 @@ async def _google(request):
     if 'error' in request.query:
         return {}
 
+    hostname = settings.settings['hostname']
+    protocol = settings.settings.get('protocol', 'http')
```
pmitros, line 93: Default protocol should be https.

learning_observer/learning_observer/google.py
```diff
@@ -114,14 +115,18 @@ def extract_parameters_from_format_string(format_string):
     return [f[1] for f in string.Formatter().parse(format_string) if f[1] is not None]
 
 
-async def raw_google_ajax(request, target_url, **kwargs):
+async def raw_google_ajax(runtime, target_url, **kwargs):
     '''
     Make an AJAX call to Google, managing auth + auth.
 
-    * request is the aiohttp request object.
+    * runtime is a Runtime class containing request information.
```
pmitros, line 122: In the future, the runtime may contain other sources of Google auth information.

learning_observer/learning_observer/google.py
```diff
@@ -114,14 +115,18 @@ def extract_parameters_from_format_string(format_string):
...
     * default_url is typically grabbed from ENDPOINTS
     * ... and we pass the named parameters
     '''
+    if isinstance(runtime, learning_observer.runtime.Runtime):
+        request = runtime.get_request()
+    else:
+        request = runtime
```
pmitros, line 129: This should raise an exception, not set `request` to `runtime`. Alternatives:

- They could both be keyword parameters (default `None`). 
- We could always pass in a runtime.

A short-term alternative would be to print a warning, so we can clean up legacy code paths where this is still `request`. 

`request = runtime` is a bug.

learning_observer/learning_observer/google.py
```diff
@@ -389,6 +403,9 @@ def extract_text_from_google_doc_json(
     within text chunks, since we do have different lengths for something like
     this flag. It does work okay globally.
     '''
+    # return error message for text
+    if 'error' in j:
+        return get_error_message(j['error'])
```
pmitros, line 408: This is very bad code. We have no way to know whether this is an error or a student text. Options:

- Define a new exception, and raise an exception here with the appropriate error messages
- Return something like a JSON structure instead

modules/writing_observer/writing_observer/aggregator.py
```diff
@@ -193,13 +175,89 @@ async def merge_with_student_data(writing_data, student_data):
...
+        Retrieves the latest documents for a set of students using Google Docs.
+
+        :param student_data: The student data.
+        :return: The latest documents.
+        """
+        writing_data = [
+            {'text': await fetch_doc_from_google(get_last_document_id(s))}
+            if get_last_document_id(s) is not None else {}
+            for s in student_data
+        ]
```
pmitros, line 238: This would actually be a logical structure for `fetch_doc_from_google` to return. `{'text':` or `{'error': `

learning_observer/learning_observer/kvs.py
```diff
@@ -286,26 +287,37 @@ def kvs_startup_check():
...
+        for kvs_item in learning_observer.settings.settings['kvs']:
+            kvs_type = kvs_item['type']
+            kvs_class = KVS_MAP[kvs_type]
+            # set timeout for ephemeral redis connections
+            if kvs_type == 'redis_ephemeral':
+                kvs_class = functools.partial(kvs_class, kvs_item['expiry'])
+            KVS_DICT[kvs_item['name']] = kvs_class
+
+        # Set default KVS to not break legacy code
+        KVS = KVS_DICT.get('main', InMemoryKVS)
```
pmitros, line 307: This should very much *not* have a default like this. 

If a deployed system falls into a default like this -- Collin's deploy at NCSU by upgrading, or a new setup, it *will* lose data on each restart. 

On code like this, you want to fail early.

learning_observer/learning_observer/cache.py
```diff
@@ -0,0 +1,27 @@
+from functools import wraps
```
pmitros, line 1: Unless there's a good reason, not: `import functools` and then `functools.wraps`

learning_observer/learning_observer/creds.yaml.example
```diff
@@ -43,15 +44,16 @@ kvs:
...
     # redis_ephemeral for redis / debugging (object disappear)
     # redis for deploy
-    type: stub
     # If using redis_ephemeral, persist objects for 60s = 1 minute
     # I typically use:
     # * 1-10s for test cases
     # * 1-5 minutes for interactive debugging
     # * 6-24 hours for development
-    expiry: 60
+  - name: main
```
pmitros, line 52: nit: should be a dictionary with defaults

learning_observer/learning_observer/google.py
```diff
@@ -368,6 +370,16 @@ def _force_text_length(text, length):
     return text[:length] + " " * (length - len(text))
 
 
+def get_error_message(error):
```
pmitros, line 373: `get_error_details`

PR #57, opened 2023-03-30
Analysis wrapper fix

bradley-erickson:
- Linted code
- Removed unused methods
- Cleaned up if statement logic


bradley-erickson:
Builds on and closes #44

PR #59, opened 2023-04-04
added redirect to login

bradley-erickson:


pmitros:
FYI: This is slightly incorrect:

- It works for HTML. JSON responses should be a bit different so the JS can handle this properly
- There should be some indication of why the user is being taken to `/` (or they should be taken to a `/login` page)
- We should encode where they came from, so upon authorization, they can go back there. 
- We should be careful about not logged in versus unauthorized; if a teacher tries to access an admin resource, that should not have this behavior. 

It probably makes sense to file an epic, and do those in a future PR. If this works correctly, you can either:
1. squash and merge
2. do an intermediate fix, where instead of "Please log in", there is a more meaningful page (e.g. link to the login page, link back, etc.)

I prefer #2, but not sure it's worth your time right now.

PR #67, opened 2023-04-10
WiP: Individual student dashboard

bradley-erickson:
Creating draft pull request for initial pieces of feedback

### DIFF
--- a/modules/lo_dash_react_components/src/lib/components/StudentSelectHeader.react.js
+++ b/modules/lo_dash_react_components/src/lib/components/StudentSelectHeader.react.js
@@ -68,38 +68,40 @@ export default class StudentSelectHeader extends Component {
 
     return (
       <div id={id} className={classNames}>
-        <header id="student-select-header" className="nav-header">
-          <div className="button-left">
-            <button className="button" onClick={() => this.iterateName(-1)}>
-              &lt;
-            </button>
-          </div>
-          <div className="header-student" onClick={this.handleClick}>
-            {selectedStudent}
-          </div>
-          {showDropdown && (
-            <ul className="dropdown">
-              {students.map((name) => (
-                <li
-                  key={name}
-                  onClick={() => this.handleSelect(name)}
-                  className={
-                    name === selectedStudent
-                      ? "dropdown-item dropdown-item-selected"
-                      : "dropdown-item"
-                  }
-                >
-                  {name}
-                </li>
-              ))}
-            </ul>
-          )}
-          <div className="button-right">
-            <button className="button" onClick={() => this.iterateName(1)}>
-              &gt;
-            </button>
-          </div>
-        </header>
+        <div id="student-select-header">
+          <header className="nav-header">
+            <div className="button-left">
+              <button className="button" onClick={() => this.iterateName(-1)}>
+                &lt;
+              </button>
+            </div>
+            <div className="header-student" onClick={this.handleClick}>
+              {selectedStudent}
+            </div>
+            {showDropdown && (
+              <ul className="dropdown">
+                {students.map((name) => (
+                  <li
+                    key={name}
+                    onClick={() => this.handleSelect(name)}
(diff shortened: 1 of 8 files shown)


modules/lo_dash_react_components/src/lib/components/StudentSelectHeader.react.js
```diff
@@ -68,38 +68,40 @@ export default class StudentSelectHeader extends Component {
...
-              ))}
-            </ul>
-          )}
-          <div className="button-right">
-            <button className="button" onClick={() => this.iterateName(1)}>
-              &gt;
-            </button>
-          </div>
-        </header>
+        <div id="student-select-header">
```
pmitros, line 71: Why the extra DOM element?

modules/wo_teacher_per_student_view/wo_teacher_per_student_view/module.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+COURSE_DASHBOARDS = [{
+    'name': "Writing Observer per student view",
+    'url': "/wo_teacher_per_student_view/dash/dashboard/",
+    "icon": {
+        "type": "fas",
+        "icon": "fa-pen-nib"
+    }
+}]
```
pmitros, line 105: We should find some way to omit repetitive cut-and-paste code like this.

bradley-erickson: Agreed.

modules/lo_dash_react_components/src/lib/components/StudentSelectHeader.react.js
bradley-erickson: A bug existed in the scss where the `nav-header` items were not being properly applied. Defining the following (how it currently is):
```scss
#student-select-header {
    .nav-header {
    ...
```
Requires `nav-header` to be a child of `student-select-header`. Prior to this PR, they were defined at the same level, thus causing `nav-header` to not use the appropriate css. The extra DOM element addresses this.

pmitros: Wouldn't the right fix be to fix the CSS? Perhaps this should be `#student-select-header {` or `.nav-header {`, but not both nested.

bradley-erickson: Probably. The reason I went for adding the extra element is so we are consistent in our scss file and have
```scss
#student-select-header {
    .nav-header {
        item1: value;
        ...
    }
    .header-student {
    }
    ...
```

We could simply put everything under `.nav-header` into `#student-select-header`, outside of other the classes and achieve the same functionality without adding the additional DOM element.
```scss
#student-select-header {
    item1: value;
    ...
    .header-student {
    }
    ...
```

In terms of overall readability and usage, I prefer the first option; however, both options offer a solution to the issue. I'm okay with either.

pmitros: I personally find the second option much more readable:

- In the second option, I can clearly see that `item: value` cascades down to `.header-student`. 
- In the first option, unless I'm familiar with the DOM structure (which I won't be even a week or two later), I'll need to do some digging to understand why `.header-student` has the additional styles. From the CSS, they look like parallel elements. 

I don't know CSS best practices here, and if you have an argument for the former, I'm glad to learn something new. If there's a logical reason you prefer the first option, please let me know (even if it's not important). If it's just preference, we should go with #2.

bradley-erickson: I adjusted this to follow a similar, but slightly different option 2. 
A few scss best practices I found online suggested avoiding referring to items by id (#). Since we already have a top level classname with the same name, I just adjusted changed the id to instead be a class selector.

PR #69, opened 2023-04-10
WiP: Communication stream

bradley-erickson:
Here is a possible implementation for the communication stream.

- Each module defines a `client_data_schema` that extends a `base_schema`
  - The `base_schema` defines generic metadata we may want from the user
- When messages are received from the user, their structure is validated
- Then passed to the aggregator for usage

### DIFF
--- a/learning_observer/learning_observer/dashboard.py
+++ b/learning_observer/learning_observer/dashboard.py
@@ -362,19 +362,32 @@ async def process_function(func, parameters, runtime, sd):
     Passes the output from one function to the next as the 'prev_output' parameter.
 
     :param func: A function or a list of functions and/or nested lists of functions.
+    :param parameters: A dictionary of key-value pairs to pass into each function
     :param runtime: The runtime object.
     :param sd: The student_data object.
     :return: A dictionary with the final output of the functions.
+
+    Example 1 - func is a function:
+    ```python
+    # inputs
+    func = add
+    parameters = {'x': 1, 'y': 2}
+    # add 1 and 2 together
+    output = process_function(func, parameters, runtime, sd)
+    # 3
+    ```
+
+    Example 2 - func is a list:
+    ```python
+    # inputs
+    func = [sub, [mult, double]]
+    # should be same nested list structure as the func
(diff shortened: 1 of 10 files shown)


learning_observer/learning_observer/dashboard.py
```diff
@@ -354,6 +356,199 @@ async def websocket_dashboard_view(request):
...
+        :param func: The function to execute.
+        :param params: The parameters to pass to the function.
+        :return: The result of the function execution.
+        """
+        result = func(**params)
+        if await is_async_function(func):
+            result = await result
+        return result
+
+    async def process_nested_function(func, params, prev_output=None):
```
pmitros, line 391: The comment should (1) give an example of usage. (2) explain why we're doing this. 

This seems like it might have a bit too much magic in it, but I'll have to see how it's used.

bradley-erickson:
To test out connecting to the websocket through this dashboard
```js
parameters={
    aggregators: [
        {'x': 1, 'y': 2},
        [
            {'x': 3, 'y': 4},
            [
                {'y': 2},
                {}
            ]
        ]
    ],
    refresh:1
}
msg = {course_id: course_id, module: 'test_module', parameters: parameters}
var ws = new WebSocket("ws://localhost:8888/wsapi/dashboard-generic");
ws.onmessage = function(event) {
  console.log(JSON.parse(event.data));
};
ws.send(JSON.stringify(msg))
```

learning_observer/learning_observer/dashboard.py
```diff
@@ -354,6 +356,259 @@ async def websocket_dashboard_view(request):
...
+                default_data
+            )
+
+        # Fetch the student data using the student_state_fetcher function
+        sd = await student_state_fetcher()
+        data = {
+            "student_data": sd   # Per-student list
+        }
+
+        # Process the aggregator functions with the received client data and add the results to the response data.
```
bradley-erickson, line 588: Here

bradley-erickson:
This PR is now irrelevant since we merged in the other communication protocol branch. This one stuck around as long as it did just to reference some pieces of the code. Perhaps we'll come back for the query validation stuff.

PR #77, opened 2023-05-16
WiP: Communication protocol

pmitros:
This is a format for how we define query, execution trees, and request data. It's a work-in-progress. Do not merge.


learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,447 @@
...
+        right_dict_match = right_dict.get(lookup_key)
+
+        if right_dict_match:
+            merged_dict = {**left_dict, **right_dict_match}
+            result.append(merged_dict)
+
+    return result
+
+
+@handler(learning_observer.communication_protocol.query.DISPATCH_MODES.MAP)
```
bradley-erickson, line 141: implemented map function below

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+import learning_observer.module_loader
+import learning_observer.settings
+import learning_observer.stream_analytics.fields
+import learning_observer.stream_analytics.helpers
+from learning_observer.communication_protocol.util import get_nested_dict_value
+from learning_observer.communication_protocol.exception import DAGExecutionException
+
+dispatch = learning_observer.communication_protocol.query.dispatch
+
+KVS = None
```
pmitros, line 25: This shouldn't be here. It should be used from `learning_observer.kvs`

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    """
+    Handler for unimplemented functions.
+
+    :param args: Positional arguments
+    :param kwargs: Keyword arguments
+    :return: A dictionary indicating unimplemented status
+    :rtype: dict
+    """
+    raise DAGExecutionException(
+        f'Unimplemented function',
```
pmitros, line 38: No fstring

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+
+
+def unimplemented_handler(*args, **kwargs):
+    """
+    Handler for unimplemented functions.
+
+    :param args: Positional arguments
+    :param kwargs: Keyword arguments
+    :return: A dictionary indicating unimplemented status
+    :rtype: dict
```
pmitros, line 35: Nothing in this docstring is useful. It's taking up air. `return` is also not correct. I'd like to know:
* What is a handler? 
* Why do we have one for unimplemented things? What's unimplemented?
* When does this get called and why?

Those are the kinds of things I'm figuring out as I'm reading this, and it's convenient if it's right here. The docstring should be a mile-high view for people unfamiliar with the code.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    :param args: The positional arguments for the function
+    :type args: list
+    :param kwargs: The keyword arguments for the function
+    :type kwargs: dict
+    :return: The result of the function call
+    """
+    function = functions[function_name]
+    try:
+        result = function(*args, **kwargs)
+        if inspect.iscoroutinefunction(function):
```
pmitros, line 79: I think there's an `isawaitable` which you can apply to the result. That'd be a bit cleaner.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+            inspect.currentframe().f_code.co_name,
+            {'function_name': function_name, 'args': args, 'kwargs': kwargs, 'error': str(e)}
+        )
+    return result
+
+
+@handler(learning_observer.communication_protocol.query.DISPATCH_MODES.PARAMETER)
+def substitute_parameter(parameter_name, parameters, required, default):
+    """
+    Substitutes a parameter from the provided parameters.
```
pmitros, line 93: Parameters are provided where, and by whom?

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+            )
+        else:
+            return default
+    return parameters[parameter_name]
+
+
+@handler(learning_observer.communication_protocol.query.DISPATCH_MODES.JOIN)
+def handle_join(left, right, left_on=None, right_on=None):
+    """
+    Joins two lists of dictionaries based on provided keys. If no keys are provided, zips the lists together.
```
pmitros, line 116: Examples are good. Most of these can be tested with `doctest`

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+                result.append(merged_dict)
+        except DAGExecutionException as e:
+            result.append(e.to_dict())
+
+    return result
+
+
+def exception_wrapper(func):
+    """
+    Wraps a function to catch any exceptions it might throw.
```
pmitros, line 156: * Why? 
* Why are we returning exceptions? At some point, don't we want error objects going back (e.g. JSON bits)? Is this for compatibility with gather? Where does this fit in the processing pipeline?

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    """
+    Asynchronously applies a coroutine function to a list of values sequentially.
+
+    :param func: The coroutine function to be applied
+    :type func: coroutine
+    :param values: The values that the function will be applied to
+    :type values: list
+    :param value_path: The path to fetch the value to be mapped
+    :type value_path: str
+    :return: The list of results from applying the coroutine function
```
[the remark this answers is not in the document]
pmitros: Depending on how you write this, the documentation could also simply reference "We call / dispatch / route the `map` for function [x], which describes [y]` so we don't repeat in all the map coroutines.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+            output = func(get_nested_dict_value(v, value_path))
+        except DAGExecutionException as e:
+            output = e.to_dict()
+        outputs.append(output)
+    return outputs
+
+
+def annotate_map_metadata(function, results, values, value_path, func_kwargs):
+    """
+    Annotates map results with metadata related to the mapping operation.
```
pmitros, line 252: I would have no idea what this means if I hadn't designed this:

* What type of metadata?
* Why do we have metadata?

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+        if parallelize:
+            results = await map_coroutine_parallelize(partial, values, value_path)
+        else:
+            results = await map_coroutine_serial(partial, values, value_path)
+    else:
+        exception_catcher = exception_wrapper(partial)
+        if parallelize:
+            results = map_parallelize(exception_catcher, values, value_path)
+        else:
+            results = map_serialize(exception_catcher, values, value_path)
```
pmitros, line 326: I don't know if this is a good idea here, but I tend to do these as lookup tables:
* I can get the function name as `function_name = f'map{"_coroutine" if coroutine else ""}_{"parallel" if parallel else "serial"}'`
* I can reference into a dictionary mapping the strings for the four functions to the function, and call the appropriate one
* I can await if it's awaitable

I'm not sure that will be more readable in this case, but that does usually tend to be more readable and explicit than complex `if` statements.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+async def handle_select(keys, fields):
+    """
+    Select data from a key-value store (KVS) based on a list of keys.
+
+    :param keys: The keys to select data for
+    :type keys: list
+    :param fields: A mapping of key paths to keys (similar to SQL `AS`), ex) `{path.to.item: output}`
+    :type fields: dict
+    :return: The selected data ex) `[{output: value, context: {}}, ...]`
+    :rtype: list
```
pmitros, line 341: docstring here could probably use an in-memory KVS to give a simple example.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+                    'context': k['context']
+                }
+            }
+        else:
+            raise DAGExecutionException(
+                f'Key not formatted correctly for select: {k}',
+                inspect.currentframe().f_code.co_name,
+                {'keys': keys, 'fields': fields}
+            )
+        kvs_out = await KVS[k['key']]
```
pmitros, line 361: I don't think I would call this `kvs_out` but rather something suggesting it's the value resulting from querying the KVS for that key.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+        fields = {}
+
+    response = []
+    for k in keys:
+        if isinstance(k, dict) and 'key' in k:
+            item = {
+                'context': {
+                    'key': k['key'],
+                    'context': k['context']
+                }
```
pmitros, line 353: In retrospect, `context` is a horrible name. It should be something like `providence` or `processing_history` or similar.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    :return: The selected data ex) `[{output: value, context: {}}, ...]`
+    :rtype: list
+    """
+    if fields is None:
+        fields = {}
+
+    response = []
+    for k in keys:
+        if isinstance(k, dict) and 'key' in k:
+            item = {
```
pmitros, line 349: `query_response_element` or something?

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    :param items: The items to generate keys for
+    :type items: list
+    :return: The generated keys
+    :rtype: list
+    """
+    pass
+
+
+@handler(learning_observer.communication_protocol.query.DISPATCH_MODES.KEYS)
+def hack_handle_keys(function, STUDENTS=None, STUDENTS_path=None, RESOURCES=None, RESOURCES_path=None):
```
pmitros, line 390: Needs a docstring explaining the hack, why we have it, and how we plan to eventually fix it.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+                'context': c,
+                'default': func['default']
+            }
+        )
+    return keys
+
+
+def _has_error(node):
+    '''
+    Non-recursive function to find and return 'error' value and its path from any dictionary within the node.
```
pmitros, line 436: I don't know:
* Why you need this and when it's called
* What an 'error' value is
* What a node is

This sort of mile-high context is important.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    '''
+    queue = [(node, [])]  # start with the node and an empty path
+    while queue:
+        current, path = queue.pop(0)
+        if 'error' in current:
+            return current, path
```
pmitros, line 442: Note this will only return one error, even if there are multiple. Of course, I have no idea right now what kinds of errors this is looking for until I understand the context in the codebase.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+        for c in current:
+            if isinstance(current[c], dict):
+                queue.append((current[c], path + [c]))
+            elif isinstance(current[c], list):
+                for idx, i in enumerate(current[c]):
+                    if isinstance(i, dict):
+                        queue.append((i, path + [c, idx]))
+    return None, []
```
pmitros, line 450: Depth-first search will be exponentially slow (in big O). This will only matter if you have a complex, interconnected DAG, but keeping some log of visited nodes would prevent this.

pmitros: This could be fixed or (since we have simple DAGs now) left as a TODO, but it should be clearly documented (with appropriate keywords if someone is looking for performance issues).

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+
+
+KEYS_TO_REMOVE = ['context']
+
+
+def _sanitaize_output(variable):
+    '''
+    Sanitizes output by removing specified keys from each level of a dictionary or a list of dictionaries.
```
pmitros, line 458: sanitize <-- spelling

Shouldn't `KEYS_TO_REMOVE` be a parameter (perhaps with a default?). Why is this considered a sanitization? Would we be leaking sensitive data here, or are we just cleaning unnecessary data?

In the docstring, it's not so much necessary to explain what this does as why we're doing it. 

If all we're using this for, the name might be `strip_context` or similar.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+        return {key: value for key, value in variable.items() if key not in KEYS_TO_REMOVE}
+    elif isinstance(variable, list):
+        return [_sanitaize_output(item) if isinstance(item, dict) else item for item in variable]
+    else:
+        return variable
+
+
+async def execute_dag(endpoint, parameters, functions, target_exports=None):
+    """
+    Execute a flattened directed acyclic graph (DAG).
```
pmitros, line 475: Explain "flattened" here. Or move the flatten function inside so it doesn't need to be flattened. This should clearly document what can and cannot go in.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+
+    :param endpoint: The endpoint dictionary
+    :type endpoint: dict
+    :param parameters: The parameters for execution
+    :type parameters: dict
+    :param functions: The functions available for execution
+    :type functions: dict
+    :return: The result of the execution
+    :rtype: dict
```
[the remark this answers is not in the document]
pmitros: I want to know the format of all the parameters to all of the functions in enough detail to be able to make use of them. As a new developer, I don't want to need to add print() statements or search the entire codebase to understand what goes in and what comes out.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    """
+    if target_exports is None:
+        target_exports = []
```
pmitros, line 487: I have no idea what this is from the code. I also don't know why we'd call this with no exports. Should this parameter really have a default of `None`? That seems like a crazy default.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    target_nodes = [endpoint['exports'][key]['returns'] for key in target_exports]
+
+    visited = set()
+    nodes = endpoint['execution_dag']
+
+    global KVS
+    if KVS is None:
+        KVS = learning_observer.kvs.KVS()
+
+    async def dispatch_node(node):
```
pmitros, line 498: Docstring

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+
+            if inspect.iscoroutinefunction(function):
+                result = await result
+            return result
+        except DAGExecutionException as e:
+            return e.to_dict()
+
+    async def walk_dict(node_dict):
+        '''
+        This will walk a dictionary, and call `visit` on all variables, and make the requisite substitions
```
pmitros, line 524: Expand docstring (What is a substitution? What does it mean to "walk a dictionary"? What does "visit" do?)

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+        '''
+        for child_key, child_value in list(node_dict.items()):
+            if isinstance(child_value, dict) and dispatch in child_value and child_value[dispatch] == learning_observer.communication_protocol.query.DISPATCH_MODES.VARIABLE:
+                node_dict[child_key] = await visit(child_value['variable_name'])
+            elif isinstance(child_value, dict):
+                await walk_dict(child_value)
+
+    async def visit(node_name):
```
pmitros, line 532: docstring

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+            nodes[node_name] = {
+                'error': error,
+                'dispatch': nodes[node_name]['dispatch'],
+                'error_path': error_path
+            }
+        else:
+            nodes[node_name] = await dispatch_node(nodes[node_name])
+
+        # import json
+        # print('*****', node_name, json.dumps(nodes[node_name], indent=2, default=str))  # useful but produces a lot
```
pmitros, line 553: Don't include commented-out code

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+        visited.add(node_name)
+        return nodes[node_name]
+
+    # return everything in dev mode
```
pmitros, line 557: 'everything' ==> something explaining that this refers to execution paths / providence. No one will know what a `context` is.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -0,0 +1,562 @@
...
+    if learning_observer.settings.RUN_MODE == learning_observer.settings.RUN_MODES.DEV:
+        return {e: await visit(e) for e in target_nodes}
+
+    # otherwise remove context from outputs
```
pmitros, line 561: `# Remove execution history if in deployed settings, with data flowing back to teacher dashboards`

PR #82, opened 2023-07-05
Arglab manifest updates

bradley-erickson:
Adding in missing arglab extension changes.


bradley-erickson:
Migrating a few changes from the arglab

PR #87, opened 2023-07-26
Language tool integration plus dashboard

bradley-erickson:


learning_observer/learning_observer/communication_protocol/util.py
```diff
@@ -62,6 +62,9 @@ def get_nested_dict_value(d, key_str=None):
     '''
     k = key_str
     output = learning_observer.util.get_nested_dict_value(d, k)
+    # FIXME if the output is None, then we still raise an error
+    # we ought to allow the output to be None without raising an error
+    # sometimes the default of a reducer might be None
```
pmitros, line 67: Fix in the original place, rather than adding a wrapper:

```
RAISE_EXCEPTION = sentinel()

def get_nested_value(d, key_str, default=RAISE_EXCEPTION):
    pass
```

learning_observer/learning_observer/util.py
```diff
@@ -167,6 +168,21 @@ def timeparse(timestamp):
     return parser.isoparse(timestamp)
 
 
+def check_service(host, port):
```
pmitros, line 171: We ought to:

- Connect on startup in the module
- If connection fails on startup, we should flag a startup error (LT not running; run it or disable the module)
- Persist the connection, with reconnects if we lose one

Connecting and disconnecting seems like the wrong way to do this.

modules/wo_common_student_errors/setup.cfg
```diff
@@ -0,0 +1,16 @@
...
+url = https://github.com/ETS-Next-Gen/writing_observer
+version = 0.1
+
+[options]
+packages = wo_common_student_errors
+include_package_data = true
+
+[options.entry_points]
+lo_modules =
+   lo_core = wo_common_student_errors.module
```
pmitros, line 13: We should both understand this :)

PR #95, opened 2023-09-19
Comm protocol improvements

bradley-erickson:
Added ability to reference execution dags in other modules.


learning_observer/learning_observer/dashboard.py
```diff
@@ -361,6 +363,33 @@ async def DAGNotFound(dag_name):
     return f'Execution DAG, {dag_name}, not found on system.'
 
 
+def find_dependent_dags(dag, deps=None):
+    '''Check if the execution DAG has any dependencies on other DAGs
```
pmitros, line 367: Doc string:

This will return any dependencies on other dags (in other words, variables of the form `namespace.variable` which are not local). If it finds something like `writing_observer.docs`, it will add that to a set, and return that set of dependencies.

This is so we can have a complete list of DAG nodes we might need to reference. We can take the keys from the `execution_dag` dictionary, but we also add all of the references to DAGs in other modules.

learning_observer/learning_observer/dashboard.py
```diff
@@ -361,6 +363,33 @@ async def DAGNotFound(dag_name):
...
+        if dag.get('dispatch') == learning_observer.communication_protocol.query.DISPATCH_MODES.VARIABLE:
+            variable_name = dag['variable_name']
+            if '.' in variable_name:
+                deps.add(variable_name.split('.')[0])
+        for key in dag:
+            find_dependent_dags(dag[key], deps)
+    return deps
+
+
+def add_prefix_to_dep_dag_variables(dag, prefix):
```
pmitros, line 382: Function name: `fully_qualify_names_with_default_namespace(dag, namespace_prefix)`

Doc string: 

In order to avoid naming conflicts, and to have consistency, this will append the default namespace to all variables without a namespace prefix. For example, if the prefix is `writing_observer`, and we run into a node `docs`, this node will be renamed to `writing_observer.docs`. If a node already has a prefix (e.g. `dynamic_assessment.foo`), it will remain unchanged.

learning_observer/learning_observer/dashboard.py
pmitros: function name: `extract_namespaced_dags`

pmitros:
Looks good to merge with new docstrings and function names. Doctests would be nice, but not a showstopper for merge. Also, do mention these should be moved out of dashboard as a todo in the comments.

PR #99, opened 2023-09-26
Starting to abstract out event code

bradley-erickson:


bradley-erickson:
I left a lot of console.log statements in for the time being while we continue to test it out.

- [ ] Initialize data properly
- [ ] Add metadata related fields - skipped implementation while we worked on the end-to-end pipeline of the code
- [ ] Clean up console.log statements
- [ ] Implement Ajax logger
- [ ] Properly handle deny (both server and client side)
- [ ] Handle sending queued events as to not overload whatever server we are sending them to
- [ ]

extension/writing-process/README.md
```diff
@@ -20,12 +20,14 @@ npm install
 
 | Commands | Description |
 | --- | --- |
-| `npm run start` | builds extension into `dist/`, watches for file changes |
-| `npm run build` | generate release version - `release.zip` |
-| `npm run docs` | generate source code docs into `public/documentation` |
-| `npm run clean` | removes `dist/` directory |
-| `npm run test` | run unit tests |
-| `npm run sync` | update projects config files for `extension-cli` |
+| `npm run ext:start` | builds extension into `dist/`, watches for file changes |
```
pmitros, line 23: Document what `ext` means here. The README is a really nice place for people to know what each set are.

extension/writing-process/src/background.js
```diff
@@ -9,9 +9,15 @@ var RAW_DEBUG = false;
 /* This variable must be manually updated to specify the server that
  * the data will be sent to.  
 */
-var WEBSOCKET_SERVER_URL = "wss://learning-observer.org/wsapi/in/" 
+// var WEBSOCKET_SERVER_URL = "wss://learning-observer.org/wsapi/in/" 
+const WEBSOCKET_SERVER_URL = 'ws://127.0.0.1:8765/ws'
```
pmitros, line 13: I might not change this here, until we figure this out. It's fine if this goes nowhere for now (and in many ways better than if it goes to `localhost`).

extension/writing-process/src/background.js
```diff
@@ -9,9 +9,15 @@ var RAW_DEBUG = false;
...
 
 import { googledocs_id_from_url } from './writing_common';
+
+import { init as loggerInit, logEvent, consoleLogger, websocketLogger, VERBS } from 'lo_event';
+
+const loggers = [consoleLogger, websocketLogger(WEBSOCKET_SERVER_URL, chrome.storage.sync)]
+loggerInit('', '', loggers, '', '')
```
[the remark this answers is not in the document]
pmitros: ` source` <-- org.mitros.writing
`version` <-- old version + 0.01
`metadata`, <-- we can set later, but needs a todo comment. Remind ourselves of initial / extra metadata (for before / after auth / opt-out)
`debugLevel` <-- set something

extension/writing-process/webpack.config.js
```diff
@@ -0,0 +1,41 @@
...
+                test: /\.css$/,
+                use: ['style-loader', 'css-loader'],
+            },
+            {
+                test: /\.html$/,
+                use: ['html-loader'],
+            }
+        ],
+    },
+    externals: {
```
pmitros, line 37: This __definitely__ needs a comment.

modules/lo_event/README.md
```diff
@@ -0,0 +1,36 @@
+# lo_event
```
pmitros, line 1: At some point, when we revisit this code, we should tag me to review / edit this text. We should do ping-pong here.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,253 @@
...
+// A list of all loggers which should receive events.
+let loggersEnabled = []
+// This can either be:
+// * dictionaries of metadata
+// * functions which return metadata
+// * asynchronous functions which return metadata
+const metadata = []
+const blocked = false
+
+export const VERBS = xapi.VERB
```
pmitros, line 34: XAPI_VERBS or xapi.verbs or similar. Personally, I'd do: `export xapi.verbs` if that works with export syntax.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,253 @@
...
+export function logEvent (eventType, event) {
+  event.event_type = eventType
+
+  if (event.wa_source === null) {
+    event.wa_source = 'background_page'
+  }
+  const jsonEncodedEvent = JSON.stringify(event)
+
+  for (let i = 0; i < loggersEnabled.length; i++) {
+    loggersEnabled[i](jsonEncodedEvent)
```
pmitros, line 59: We should think through if we will ever need `await` here (and in other places)

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,253 @@
...
+    return socket
+  }
+
+  socket = newWebsocket()
+
+  function areWeDone () {
+    console.log('are we done yet?')
+    if (state.has('chrome_identity') &&
+      state.has('local_storage') &&
+      state.has('allow') &&
```
pmitros, line 105: `allow` and `deny` should be abstracted out, since it's common to many loggers.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,253 @@
...
+      state.add('chrome_identity')
+      areWeDone()
+    })
+    const storageKeys = [
+      'teacher_tag', // Unused. In the future: whom do we authorize to receive data?
+      'user_tag', // Unique per-user tag in the settings page
+      'process_server', // Unused. Which server should we send to?
+      'unique_id', // Unique ID set in the settings page
+      'generated_id', // Autogenerated, for future forensics
+      'opt_out'
```
pmitros, line 140: This stuff is specific to Writing Observer and not generic. We'll want to move this to be passed into init.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,253 @@
...
+    // Send the server the user info. This might not always be available.
+    state = new Set()
+    let event
+    profileInfoWrapper().then((userInfo) => {
+      event = {
+        chrome_identity: userInfo
+      }
+      socket.send(JSON.stringify(event))
+      console.log('resolving profileInfoWrapper', event)
+      state.add('chrome_identity')
```
pmitros, line 131: A lot of the auth stuff is specific to being in an extension. Again, we'll want to tease this out.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,253 @@
...
+        console.log(event)
+        // event = add_event_metadata('local_storage', event)
+        socket.send(JSON.stringify(event)) // do we need this to send right here, we aren't fully ready
+        state.add('local_storage')
+        areWeDone()
+      }
+    })
+  }
+
+  function receiveMessage (event) {
```
pmitros, line 168: We'll want some kind of dispatch here on the event type. This will not work very well as an interface if we ever need more back-and-forth. We should expect a JSON object, and handle either auth / opt-out, or let the server know it's an unsupported event.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,253 @@
...
+
+  return function (data) {
+    console.log('queue data', data)
+    queue.enqueue(data)
+    dequeue()
+  }
+}
+
+// storage mirrors the capability of `chrome.storage.sync`
+// this is used for testing purposes
```
pmitros, line 224: "and in node"

Should we use our sqlite / indexeddb, or similar?

modules/lo_event/queue.cjs
```diff
@@ -0,0 +1,205 @@
+/*
```
pmitros, line 1: Remove this file please

modules/lo_event/tests/queue.test.js
```diff
@@ -0,0 +1,18 @@
+import { Queue } from '../lo_event/queue.js'
```
pmitros, line 1: Comment about where we are w.r.t. in the test framework.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,163 @@
...
+          if (error instanceof blacklist.BlockError) {
+            // Handle BlockError exception here
+            blacklist.handleBlockError(error);
+          } else {
+            // Other types of exceptions will propagate up
+            throw error;
+          }
+        }
+      }
+      await dequeue();
```
pmitros, line 132: May run us out of stack space. This should not be recursive. The loop before was not quite right, but we can either:
* Chain functions (e.g. `.then()`)
* Make a for loop which is correct

As we work through this, we might want to think about what happens if multiple things call dequeue while a dequeue is already running. The key thing is we only want one scheduled at a time.

modules/lo_event/lo_event/queue.js
```diff
@@ -103,21 +103,42 @@ export class Queue {
...
+  }
+
+  dbDequeue () {
     return new Promise((resolve, reject) => {
       if (this.db === null) {
-        if (this.memoryQueue.length > 0) {
-          resolve(this.memoryQueue.shift());
-        }
-        reject(new Error('db is null and memory queue is empty'));
+        resolve(null);
```
pmitros, line 141: This really needs to be documented.

pmitros: `await databaseIsReady()` before doing any transactions

modules/lo_event/lo_event/queue.js
```diff
@@ -35,46 +38,40 @@ export class Queue {
...
-        await delay(1000);
-        this.dbEnqueue();
-      })();
-      return null;
-    }
-
-    // Enqueue the next object
-    const transaction = this.db.transaction([this.queueName], 'readwrite');
-    const objectStore = transaction.objectStore(this.queueName);
+    backoff(() => this.db !== null & this.resolveDBReady === null, 'Database never got ready').then(() => {
```
bradley-erickson, line 41: Give then entirety of the contents of .then into its own named method that describes what it does.

modules/lo_event/lo_event/util.js
```diff
@@ -286,3 +286,18 @@ export async function mergeMetadata (inputList) {
 export function delay (ms) {
   return new Promise(resolve => setTimeout(resolve, ms));
 }
+
+export async function backoff (
+  predicate,
+  errorMessage = 'Could not resolve backoff function',
+  delays = [100, 1000, 1000 * 10, 1000 * 60, 1000 * 60 * 5, 1000 * 60 * 30]
```
pmitros, line 293: // milliseconds, defaults to 100ms through 30m before failing

pmitros: I'm not sure it's worth the extra lines since this is in just one place, but I often do:

MS = 1
SECS = 1000*MS
...

And that way, I can write 100*MS, 30*MINS, etc. This is nice for calling code too, since it avoids the bug of where 1000 can mean 1 second or 16 minutes depending on which sleep is called.

modules/lo_event/lo_event/util.js
```diff
@@ -286,3 +286,18 @@ export async function mergeMetadata (inputList) {
...
+) {
+  for (let i = 0; i < delays.length; i++) {
+    if (await predicate()) {
+      return;
+    }
+    console.log('waiting for ', delays[i]);
+    await delay(delays[i]);
+  }
+  throw new Error(errorMessage);
+}
```
pmitros, line 303: Termination condition could be specified (die, versus continue repeating every week).

modules/lo_event/lo_event/queue.js
```diff
@@ -218,11 +223,18 @@ export class Queue {
...
-  count () {
+  async count () {
     if (this.db === null) {
       return this.memoryQueue.length;
     }
 
+    try {
+      await backoff(() => this.resolveDBReady === null, 'Database never got ready');
+    } catch (error) {
+      debug.error('Backoff error occured', error);
```
pmitros, line 234: Backoff error doing what? Where am I? Who am I? Why am I here?

modules/lo_event/lo_event/queue.js
```diff
@@ -178,7 +177,13 @@ export class Queue {
    *
    * @returns a Promise to fetch the next available item in the db
    */
-  dbDequeue () {
+  async dbDequeue () {
+    try {
+      await backoff(() => this.resolveDBReady === null, 'Database never got ready');
+    } catch (error) {
+      debug.error('Backoff error occured', error);
```
pmitros, line 184: See below. Needs one more word

modules/lo_event/lo_event/queue.js
```diff
@@ -35,46 +38,40 @@ export class Queue {
...
+          // TODO: Test this works.
+          // For background, search for: immediately invoked async function expression
+          (async () => {
+            await delay(1000);
+            this.dbEnqueue();
+          })();
+        }
+      };
+    }).catch(error => {
+      console.error('error in backoff', error);
```
pmitros, line 73: See below

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -38,7 +38,13 @@ export function websocketLogger (server) {
     const event = { local_storage: {} };
     console.log(event);
     if (!firstConnection) {
-      queue.prepend(metadata);
+      profileInfoWrapper().then((result) => {
```
pmitros, line 41: Include a comment:

```
// This code is wrong.
// It is scaffolding as we are changing how we handle lock_fields, metadata, etc. for backwards compatibility with the old way. 
// It should be removed as soon as the server-side code is updated and we have a new handshake protocol.
// In particular, profileInfoWrapper should not be here, since that was abstracted up many levels.  
```

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -38,7 +38,13 @@ export function websocketLogger (server) {
...
+        if (Object.keys(result).length > 0) {
+          queue.prepend(JSON.stringify({ event: 'metadata_finished' }));
+          queue.prepend(JSON.stringify({ event: 'chrome_identity', chrome_identity: result }));
```
pmitros, line 44: Used outside of an extension, I won't have `chrome_identity`

extension/writing-process/src/background.js
```diff
@@ -12,242 +12,21 @@ var RAW_DEBUG = false;
...
-            var event = { "issue": "Lost connection", "code": event.code };
-            event = add_event_metadata("warning", event);
-            event = JSON.stringify(event);
-            queue.push(event);
-        };
-        return socket;
-    }
+const loggers = [
+  consoleLogger(),
+  websocketLogger(WEBSOCKET_SERVER_URL)
```
pmitros, line 24: What we would like to support is:

```
callback = new Promise((resolve, reject) => {storage.getItem('server').then((data) => resolve(data)} /* and other async logic */);

websocketLogger( callback );
```

I'm not sure if this should be done inside `websocketLogger`'s `init` (which is easy enough) or one level up.

pmitros: Perhaps document this for now, and we can figure it out more fully later.

extension/writing-process/src/background.js
```diff
@@ -12,242 +12,21 @@ var RAW_DEBUG = false;
...
-        event['wa_source'] = 'background_page';
-    }
-    var json_encoded_event = JSON.stringify(event);
-
-    for (var i=0; i<loggers_enabled.length; i++) {
-        loggers_enabled[i](json_encoded_event);
-    }
-}
+loEvent.init('org.mitros.writing_analytics', '0.01', loggers, loEventDebug.LEVEL.SIMPLE);
+loEvent.setFieldSet([loEventUtils.getBrowserInfo(), await loEventUtils.debuggingMetadata()]);
```
pmitros, line 28: Confirm `setFieldSet` can handle promises, and if so, kill the `await`.

pmitros: (And if not, let's talk again)

extension/writing-process/src/inject.js
```diff
@@ -5,16 +5,7 @@
...
-
-let result = script.textContent
-
-window.postMessage({ from: 'inject.js', data: result })
-
-script.remove()
+(function() {
+    const result = JSON.stringify(_docs_flag_initialData.info_params.token);
+    window.postMessage({ from: 'inject.js', data: result });
+})();
```
pmitros, line 11: We both need to understand whether and why we need lines eight and eleven.

pmitros: Understand what's going on here!

learning_observer/learning_observer/incoming_student_event.py
```diff
@@ -342,14 +342,15 @@ async def incoming_websocket_handler(request):
         header_events.append(json_msg)
 
     first_event = header_events[0]
-    event_metadata['source'] = first_event['source']
+
+    any(event_metadata.update(event['fields']) for event in header_events if event['event'] == 'lock_fields')
```
pmitros, line 346: Make this into a normal loop:
```
for event in header_events:
   if event['event'] == 'lock_fields':
     event_metadata.update(event['fields'])
```

extension/writing-process/src/inject.js
bradley-erickson: We need figure out:
1. Whether we need this file
2. If we can move this to an existing file or not
3. Document why we need the `_docs_flag_initialData`

modules/lo_event/lo_event/browserStorage.js
```diff
@@ -0,0 +1,104 @@
+// Most of this code is not finished. It outlines the general planned
+// logic, but we have no idea how much of this works.
```
pmitros, line 2: If you believe this code works, you should update this comment.

If you only tested in some context, this update should be precise (e.g. we know this works in the extension and in-browser but not node, or whatever it is).

We should add test cases for all of this to our to do list, if it's not already there. This code will be very hard to maintain without those.

modules/lo_event/lo_event/browserStorage.js
```diff
@@ -0,0 +1,104 @@
...
+        result[keys] = this.data[keys];
+      }
+    } else {
+      result = { ...this.data };
+    }
+    if (callback) callback(result);
+  }
+};
+
+function getWithCallback (getItem) {
```
pmitros, line 45: docstring

modules/lo_event/lo_event/browserStorage.js
```diff
@@ -0,0 +1,104 @@
...
+    const results = {};
+    for (const item of items) {
+      results[item] = getItem(item);
+    }
+    callback(results);
+  }
+  return get;
+}
+
+function setWithCallback (setItem) {
```
pmitros, line 59: docstring

modules/lo_event/lo_event/browserStorage.js
```diff
@@ -0,0 +1,104 @@
...
+// We copy the chrome.storage API since it is the most annoying,
+// being asynchronous. It's not possible to implement the other APIs
+// on top of it, but the reverse is simple.
+//
+// At the end, we export one storage object, with a consistent API,
+// using our most reliable storage.
+import * as debug from './debugLog.js';
+
+// thunkStorage mirrors the capability of `chrome.storage.sync`
+// this is used mostly for testing purposes
```
pmitros, line 19: Expand comment with something to the effect of:

... and as a fallback if normal chrome storage fails for whatever reason. We do not know, but we expect it may fail with e.g. differences in browser / extension / etc. permissions, future privacy updates, etc.

But better writing :)

modules/lo_event/lo_event/browserStorage.js
```diff
@@ -0,0 +1,104 @@
...
+  // Add compatibility modifications for window.localStorage
+  debug.info('Setting storage to window.localStorage');
+  window.localStorage.get = getWithCallback(window.localStorage.getItem);
+  window.localStorage.set = setWithCallback(window.localStorage.setItem);
+  storage = window.localStorage;
+} else {
+  // If none of the above options exist, fall back to thunkStorage or exit gracefully
+  debug.info('Setting storage to default, thunkStorage');
+  storage = thunkStorage;
+}
```
pmitros, line 104: Do a code review on this file, please.

modules/lo_event/lo_event/debugLog.js
```diff
@@ -0,0 +1,81 @@
+function sendToServer (sendEvent) {
+  const counts = {};
+  return function (messageType, message, stackTrace) {
+    if (!Object.prototype.hasOwnProperty.call(counts, messageType)) {
+      counts[messageType] = 0;
+    }
+    counts[messageType]++;
+    if (Math.log10(counts[messageType]) % 1 === 0) {
```
pmitros, line 8: Add comment:

// We confirmed that this does not give roundoff errors in Firefox and [check in Chrome], but gives exact integer answers for powers of 10

modules/lo_event/lo_event/debugLog.js
```diff
@@ -0,0 +1,81 @@
+function sendToServer (sendEvent) {
```
pmitros, line 1: This is not `sendToServer`. We might have e.g. a `redux` logger, a `console` logger, or whatnot. Better name. 

docstring too.

modules/lo_event/lo_event/debugLog.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function sendToConsole (messageType, message, stackTrace) {
+  const stackOutput = stackTrace ? `\n  Stacktrace: ${stackTrace}` : '';
+  console.log(`${messageType}, ${message} ${stackOutput}`);
+}
+
+export const DESTINATIONS = {
+  CONSOLE: sendToConsole,
+  SERVER: sendToServer
```
pmitros, line 25: `SERVER` should probably be `LOG_EVENT` or something to imply "standard `lo_event` pipeline"

modules/lo_event/lo_event/debugLog.js
```diff
@@ -0,0 +1,81 @@
...
+let debugDestinations = [DESTINATIONS.CONSOLE];
+
+export function setLevel (level) {
+  if (![LEVEL.NONE, LEVEL.SIMPLE, LEVEL.EXTENDED].includes(level)) {
+    throw new Error(`Invalid debug log type ${level}`);
+  }
+  debugLevel = level;
+}
+
+export function setDestinations (dst) {
```
pmitros, line 45: If you come up with a better name than 'destinations', I'll be happy. If not, document that this sets where debug logs go to.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,163 @@
...
+
+  try {
+    await Promise.all(initializedLoggers);
+    debug.info('Loggers initialized!');
+    initialized = INIT_LOGGERS_READY;
+  } catch (error) {
+    initialized = INIT_ERROR;
+    debug.error('Error resolving logger initializers:', error);
+  }
+  return Promise.all(initializedLoggers);
```
pmitros, line 50: Why is this here twice? Should we store this in a variable and return it above?

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,163 @@
...
+}
+
+export function setFieldSet (data) {
```
pmitros, line 53: docstring for why this is here. We should also think about how this interacts with INIT_READY

pmitros: Consider the use case of using `setFieldSet` after `INIT_READY`. I think we said we might not support this right now, but we should figure out if this works. 

Assume we were initialized eons ago. If we do `setFieldSet` `logEvent` `logEvent` in that order, are we guaranteed events are sent in the same order? If not, we should either (1) document that this is an unsupported use case; or (2) fix it.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,163 @@
...
+  const authpromises = loggersEnabled
+    .filter(logger => typeof logger.setField === 'function')
+    .map(logger => logger.setField(JSON.stringify(payload)));
+
+  await Promise.all(authpromises);
+}
+
+// TODO: We should consider specifying a set of verbs, nouns, etc. we
+// might use, and outlining what can be expected in the protocol
+export function init (
```
pmitros, line 71: To think about: We should consider structing / destructing here, in a future version.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }
+  }
+}
+
+async function dequeue () {
+  if (!isInitialized()) {
+    debug.info('failure to dequeue, not initialized');
+    return;
+  }
```
pmitros, line 125: Does the blocker work right now, if we are streaming? Shouldn't this go inside the `while` loop?

modules/lo_event/lo_event/lo_event.js
```diff
@@ -0,0 +1,163 @@
...
+  if (!disabler.storeEvents()) {
+    return;
+  }
+  event.event = eventType;
+  timestampEvent(event);
+
+  queue.enqueue(event);
+}
+
+export function logXAPILite (
```
pmitros, line 151: We would like to be able to log events roughly following xAPI conventions (and possibly, Caliper conventions). This allows us to explicitly structure events with the same fields as xAPI, and might have validation logic in the future. However, we have not figured out the best way to do this, so please treat this as stub / in-progress code. 

In the long term, we'd like to be as close to standards as reasonable.

modules/lo_event/lo_event/util.js
```diff
@@ -0,0 +1,314 @@
+import * as crypto from 'crypto';
+import { storage } from './browserStorage.js';
+
+/*
+  Browser information object, primarily for debugging. Note that not
+  all fields will be available in all browsers and contexts. If not
+  available, it will return blanks (this is even usable in node.js,
```
pmitros, line 7: not "blanks" but "null"

modules/lo_event/lo_event/util.js
```diff
@@ -0,0 +1,314 @@
...
+
+/*
+  Create a fully-qualified web socket URL.
+
+  All parameters are optional when running on a web page. On an extension,
+  we need, at least, the base server.
+
+  This will:
+  * Convert relative URLs into fully-qualified ones, if necessary
+  * Convert HTTP/HTTPS URLs into WS/WSS ones, if necessary
```
pmitros, line 181: Examples would be helpful here. Those could also be made into test cases, which would be helpful as well.

modules/lo_event/lo_event/util.js
```diff
@@ -0,0 +1,314 @@
...
+  event.metadata.ts = Date.now();
+  event.metadata.human_ts = Date();
+  event.metadata.iso_ts = new Date().toISOString();
+}
+
+/*
+  We include some metadata. Much of this is primarily for
+  debugging. For example, it's helpful to have multiple timestamps
+  in order to understand timezone issues, misset clocks, etc.
+ */
```
pmitros, line 218: This docstring appears obsolete, and somehow, this is split between timestampEvent and debuggingMetadata.

modules/lo_event/lo_event/util.js
```diff
@@ -0,0 +1,314 @@
...
+      }
+      if (result.name) {
+        metadata.name = result.name;
+      }
+      resolve(metadata);
+    });
+  });
+}
+
+function mergeDictionary (target, source) {
```
pmitros, line 238: Docstring with example and/or test case.

modules/lo_event/lo_event/util.js
```diff
@@ -0,0 +1,314 @@
...
+
+/**
+ * Merges the output of dictionaries, sync functions, and async
+ * functions into a single master dictionary.
+ *
+ * Functions and async functions should return dictionaries.
+ *
+ * @param {Array} inputList - List of dictionaries, sync functions, and async functions
+ * @returns {Promise<Object>} - A Promise that resolves to the compiled master dictionary
+ */
```
pmitros, line 256: Example usage and output

modules/lo_event/lo_event/util.js
```diff
@@ -0,0 +1,314 @@
...
+    browserInfo.window = null;
+  } else {
+    browserInfo.window = {};
+    for (let i = 0; i < windowFields.length; i++) {
+      browserInfo.window[windowFields[i]] = window[windowFields[i]];
+    }
+  }
+
+  return { browser_info: browserInfo };
+}
```
pmitros, line 119: A lot of repetition, but not clear if this can be made better. Think for ≈5 minutes.

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -0,0 +1,143 @@
...
+      case 'blocklist':
+        debug.info('Received block error from server');
+        blockerror = new BlockError(
+          response.message,
+          response.time_limit,
+          response.action
+        );
+        break;
+      default:
+        console.log('auth has not yet occured');
```
pmitros, line 78: We have no idea if this is the case. This should just be console.log('unrecognized message: ${response}`);

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -0,0 +1,143 @@
...
+import { BlockError } from './disabler.js';
+import { delay, profileInfoWrapper } from './util.js';
+import * as debug from './debugLog.js';
+
+export function websocketLogger (server) {
+  /*
+     Log to web socket server.
+  */
+  let socket;
+  let WS;
```
pmitros, line 11: Comment:

`// Websocket library. This is a variable, since it comes from different places from node than in-browser`

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -0,0 +1,143 @@
+import { Queue } from './queue.js';
+import { BlockError } from './disabler.js';
+import { delay, profileInfoWrapper } from './util.js';
+import * as debug from './debugLog.js';
+
+export function websocketLogger (server) {
+  /*
```
pmitros, line 7: Useless docstring.

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -0,0 +1,143 @@
...
+     Log to web socket server.
+  */
+  let socket;
```
pmitros, line 10: // The socket we use to send / receive data. Contrast to `ws` below (the library)

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -0,0 +1,143 @@
...
+  let metadata = {};
+
+  function newWebsocket () {
+    socket = new WS(server);
+    socket.onopen = prepareSocket;
+    socket.onerror = function (e) {
+      debug.error('Could not connect to websocket', e);
+      let event = { issue: 'Could not connect' };
+      event = JSON.stringify(event);
+      queue.enqueue(event);
```
pmitros, line 24: This looks like a guaranteed infinite loop if there's a bug elsewhere. We have the exponential back-off logic in our debug logging code, and we need the same or similar here. Logging this once, or even ten times is okay. Logging this forever is not.

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -0,0 +1,143 @@
...
+    };
+    socket.onclose = function (e) {
+      debug.error('Lost connection to websocket', e);
+      let event = { issue: 'Lost connection', code: e.code };
+      event = JSON.stringify(event);
+      queue.enqueue(event);
```
pmitros, line 30: Ditto.

pmitros: <-- I'm glad to discuss this if unclear. This should be fixed (or explained) before merge.

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -0,0 +1,143 @@
...
+           * is updated and we have a new handshake protocol.
+          */
+          queue.prepend(JSON.stringify({ event: 'metadata_finished' }));
+          queue.prepend(JSON.stringify({ event: 'chrome_identity', chrome_identity: result }));
+
+          queue.prepend(JSON.stringify(metadata));
+        }
+      });
+    } else {
+      firstConnection = false;
```
pmitros, line 59: Shouldn't this be sent on future connections too? Why is this only for the first connection?

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -0,0 +1,143 @@
...
+import * as debug from './debugLog.js';
+
+export function websocketLogger (server) {
+  /*
+     Log to web socket server.
+  */
+  let socket;
+  let WS;
+  const queue = new Queue('websocketLogger');
+  let blockerror;
```
pmitros, line 13: // Exception, if we're blacklisted. We generate this when we receive a message, which is not a helpful place to raise the exception from, so we keep this around until we're called from the client.

learning_observer/learning_observer/incoming_student_event.py
```diff
@@ -369,13 +379,35 @@ async def process_queue():
         '''We use this queue to process events once the
         system is ready to handle them.
         '''
+        # wait for the system to be ready to process events
+        try:
+            await asyncio.wait_for(ready_to_process, 10)
```
pmitros, line 404: 60 seconds

learning_observer/learning_observer/blacklist.py
```diff
@@ -0,0 +1,45 @@
...
+    pass
+
+class TIME_LIMITS:
+    pass
+
+
+action_modes = ['TRANSMIT', 'MAINTAIN', 'DROP']
+[setattr(ACTIONS, action, action) for action in action_modes]
+time_limits = ['PERMANENT', 'MINUTES', 'DAYS']
+[setattr(TIME_LIMITS, limit, limit) for limit in time_limits]
```
pmitros, line 11: Ugly. 

Color = Enum('Color', ['RED', 'GREEN', 'BLUE'])

learning_observer/learning_observer/blacklist.py
```diff
@@ -0,0 +1,45 @@
...
+        message
+    }`
+    '''
+    # TODO add in the logic to check our blacklist
+    if False:
+        # tell client to keep messages around
+        return {
+            'status': 'blocklist',
+            'action': ACTIONS.MAINTAIN,
+            'time_limit': TIME_LIMITS.MINUTES,
```
pmitros, line 30: Note that these can (and probably usually will) be numbers too.

learning_observer/learning_observer/incoming_student_event.py
```diff
@@ -340,6 +341,10 @@ async def handle_auth_events(events):
         event, we check for auth on an individual event wrapped
         in a list. This workflow feels a little weird. We should
         re-evaluate the auth code.
+        This code checks every event for auth info and will throw
+        an error on 99% of them since auth isn't present. This makes
+        for some confusing debug_log output since we'll see a lot
+        of unauthorized messages.
```
pmitros, line 348: This would be confusing! It should be fixed.

extension/writing-process/src/inject.js
pmitros: @bradley-erickson  ping

learning_observer/learning_observer/blacklist.py
pmitros: @bradley-erickson  Ping

learning_observer/learning_observer/incoming_student_event.py
```diff
@@ -269,142 +270,152 @@ def decode_and_log_event(msg):
...
+            if not authenticated:
+                try:
+                    authenticated = await learning_observer.auth.events.authenticate(
+                        request=request,
+                        headers=[event],
+                        first_event={},
+                        source=''
+                    )
+                    await update_event_handler()
+                except aiohttp.web.HTTPUnauthorized:
```
pmitros, line 359: This should not be an exception anymore, perhaps. This is an expected behavior.

Issue #84, opened 2023-07-10
Communication Protocol Second pass


bradley-erickson:
We've taken an initial pass at the communication protocol. Anything found to be bugged, not yet implemented, or any other items please leave here.

Starting with some of the TODO in the codebase:

- [ ] Verify structure of request from client before passing to Communication Protocol
- [ ] Allow some way of specifying the server sleep time
- [ ] General LO traceback serialization/deserialization
- [ ] Implement the proper handle_keys (replace `handle_hack_keys`)
- [ ] Pass KVS to internal Select (`handle_select`)
- [ ] Catch all errors where appropriate (mainly for the `get_nested_dict_value`)
- [ ] Replace `_has_error` search algorithm (currently DFS)
- [ ] Determine appropriate naming scheme in and out (the user says they want `[x, y]` from `[wo]`, they probably expect to see `wo.x` and wo.y` after; however, we actually see `wo.x_endpoint` returned)
- [ ] Clean up Dash code
- [ ] Create JS query converter

bradley-erickson:
It would be nice to have some form of query builder interface that allows users to drag n' drop and connect nodes. After their created query is validated, they should be able to submit their query with options for parameters. Import/export would be cool.
This [example uses react-diagrams](https://github.com/projectstorm/react-diagrams/tree/master/diagrams-demo-gallery/demos/demo-drag-and-drop) essentially does what we want.

bradley-erickson:
Based on a conversation in #100, we ought to have better formatting when we write the queries.

Per Piotr:
- Formatting (so we know clearly what goes into what)
- Comments
- Some kind of doctext-style thing would be nice too

```
'single_lt_combined: { 'query': <<existing stuff>>, 'description': "A long text about what it does", <<perhaps other stuff, such as inputs, outputs, etc.>> }
```

Additionally, we think through queries which can be parallel or nonparallel. We'd like the same query:

do_some_nlp('bob') as well as do_some_nlp(all_my_students())

bradley-erickson:
- [ ] Set nodes to output a static value (set once on initial load), dynamic value (consistently receive data), or memoized value (one we keep for a bit). Currently the system is mostly in the `dynamic` state with some memoized functions.

bradley-erickson:
- [ ] If part of the execution dag is not formed correctly or missing, provide better errors. (setting `"unused_node": ""` breaks the code, but no reasonable error provided). **Low hanging fruit**

[...]

PR #100, opened 2023-10-09
added tagging of documents via their titles

bradley-erickson:
Not the cleanest implementation, but given a course id and a tag, we can get a list of students. Each student will have a list of document text. Ideally this should be a document with more metadata, but for now its just text.

This has some improvements that ought to just be added to the communication protocol, such as the unwind and the group_by function. I tried to keep implementations mostly generic so we could migrate them into their later, but there is some hardcoded stuff.

### DIFF
--- a/modules/writing_observer/writing_observer/module.py
+++ b/modules/writing_observer/writing_observer/module.py
@@ -57,6 +57,7 @@ EXECUTION_DAG = {
         'overall_lt': languagetool(texts=q.variable('docs')),
         'lt_combined': q.join(LEFT=q.variable('overall_lt'), LEFT_ON='provenance.provenance.STUDENT.value.user_id', RIGHT=q.variable('roster'), RIGHT_ON='user_id'),
 
+        'latest_doc_ids': q.join(LEFT=q.variable('roster'), RIGHT=q.variable('doc_ids'), LEFT_ON='user_id', RIGHT_ON='provenance.provenance.value.user_id'),
         # the following nodes are used to fetch a set of documents' metadata based on a given tag
         # HACK: this could be a lot fewer nodes with some form of filter functionality
         #  e.g. once we get the list of documents that match the inputted tag, we just filter
@@ -108,6 +109,10 @@ EXECUTION_DAG = {
         'assignment_docs': {
             'returns': 'assignment_docs',
             'parameters': ['course_id', 'assignment_id']
+        },
+        'latest_doc_ids': {
+            'returns': 'latest_doc_ids',
+            'parameters': ['course_id']
         }
     },
 }
@@ -172,7 +177,8 @@ REDUCERS = [
     {
         'context': "org.mitros.writing_analytics",
         'scope': writing_observer.writing_analysis.student_scope,
-        'function': writing_observer.writing_analysis.document_list
+        'function': writing_observer.writing_analysis.document_list,
+        'default': {'docs': []}
     },
     {
         'context': "org.mitros.writing_analytics",
--- a/modules/writing_observer/writing_observer/module.py
+++ b/modules/writing_observer/writing_observer/module.py
@@ -31,9 +31,10 @@ process_texts = q.call('writing_observer.process_texts')
 determine_activity = q.call('writing_observer.activity_map')
 languagetool = q.call('writing_observer.languagetool')
 update_via_google = q.call('writing_observer.update_reconstruct_with_google_api')
+assignment_documents = q.call('google.fetch_assignment_docs')
 
 unwind = q.call('unwind')
-group_by = q.call('group_by')
+group_docs_by = q.call('writing_observer.group_docs_by')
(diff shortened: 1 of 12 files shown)


modules/writing_observer/writing_observer/tag_docs.py
```diff
@@ -0,0 +1,74 @@
...
+    Objects can be a single object or a list of objects.
+
+    :param keys_to_keep: List of keys to return. If `None`, all keys are returned except the head of `value_path`
+    :default keys_to_keep: None
+    :type keys_to_keep: list
+
+    TODO this ought to be query command, I'm just writing it as a function for now
+    but I'll try to keep it generic to be slid in later.
+    '''
+    items = [objects] if type(objects) != list else objects
```
pmitros, line 30: `type(objects) != list` is bad code. You would want to check if this is list-like (keeping in mind a dictionary is iterable and will iterating over it will return its keys)

modules/writing_observer/writing_observer/tag_docs.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Objects can be a single object or a list of objects.
+
+    :param keys_to_keep: List of keys to return. If `None`, all keys are returned except the head of `value_path`
+    :default keys_to_keep: None
+    :type keys_to_keep: list
+
+    TODO this ought to be query command, I'm just writing it as a function for now
+    but I'll try to keep it generic to be slid in later.
+    '''
```
pmitros, line 29: This comment could probably use a doctest so that:

1. We can clearly see what this does
2. We can test it if changing the function

modules/writing_observer/writing_observer/tag_docs.py
```diff
@@ -0,0 +1,74 @@
...
+    items = [objects] if type(objects) != list else objects
+    remove_key = value_path.split('.')[0]
```
pmitros, line 31: I might make a helper function and call it, just so this has a name.

modules/writing_observer/writing_observer/tag_docs.py
```diff
@@ -0,0 +1,74 @@
...
+            # Perhaps we need a function similar to that of get_nested_dict_value that removes them instead
+            if remove_key in new:
+                del new[remove_key]
+            new[new_name] = value
+            unpacked.append(new)
+    return unpacked
+
+
+@learning_observer.communication_protocol.integration.publish_function('group_by')
+def group_by(items, value_path):
```
pmitros, line 48: `group_xxx_by`, to avoid confusion with a generic `group_by` (or make it generic)

modules/writing_observer/writing_observer/writing_analysis.py
```diff
@@ -143,22 +144,52 @@ async def event_count(event, internal_state):
     return state, state
 
 
+@kvs_pipeline(scope=student_scope, null_state={'tags': {}})
+async def document_tagging(event, internal_state):
```
pmitros, line 148: Issues:
1. We don't remove tags. Typing #hello gives #h, #he, #hel, ... 
2. Because this is in a single dictionary, there may be performance issues relative to storing in a key scoped to a tag.

pmitros: Resolutions to above:
1. Monitor old title and new title
2. Have a tag+student scope.

modules/writing_observer/writing_observer/module.py
```diff
@@ -51,6 +55,12 @@
         'single_lt_combined': q.join(LEFT=q.variable('single_student_lt'), LEFT_ON='provenance.provenance.STUDENT.value.user_id', RIGHT=q.variable('roster'), RIGHT_ON='user_id'),
         'overall_lt': languagetool(texts=q.variable('docs')),
         'lt_combined': q.join(LEFT=q.variable('overall_lt'), LEFT_ON='provenance.provenance.STUDENT.value.user_id', RIGHT=q.variable('roster'), RIGHT_ON='user_id'),
+
```
pmitros, line 58: * Formatting (so we know clearly what goes into what)
* Comments
* Some kind of doctext-style thing would be nice too

pmitros: We could support a dictionary for children, so we can do discovery, as well as expand these nodes.
```
'single_lt_combined: {
  'query': <<existing stuff>>,
  'description': "A long text about what it does",
  <<perhaps other stuff, such as inputs, outputs, etc.>>
}
```

pmitros: Think through queries which can be parallel or nonparallel. We'd like the same query:

`do_some_nlp('bob')` as well as `do_some_nlp(all_my_students())`

bradley-erickson: Added this to the ongoing improvements sections for the communication protocol #84

PR #112, opened 2024-01-04
added document timestamps and doc source option for communication pro…

bradley-erickson:
Added reducer for tracking when documents are accessed each second. Also added initial ability to select where we want to source documents from.

### DIFF
--- a/modules/writing_observer/writing_observer/document_timestamps.py
+++ b/modules/writing_observer/writing_observer/document_timestamps.py
@@ -1,0 +1,37 @@
+import bisect
+
+import learning_observer.communication_protocol.integration
+
+
+@learning_observer.communication_protocol.integration.publish_function('source_selector')
+def select_source(sources, source):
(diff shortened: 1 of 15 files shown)


modules/writing_observer/writing_observer/document_timestamps.py
```diff
@@ -0,0 +1,46 @@
...
+    in reference to the `requested_timestamp`.
+
+    `requested_timestamp` should be a string of ms since unix epoch
+    '''
+    output = []
+    for student in overall_timestamps:
+        timestamps = student.get('timestamps', {})
+        student['doc_id'] = ''
+        if requested_timestamp is None:
+            # perhpas this should fetch the latest doc id instead
```
pmitros, line 35: Typo

pmitros:
Looks good to me and ready to merge. 

This was simpler than I expected. :)

PR #114, opened 2024-01-09
added constants file

bradley-erickson:
Added initial set of constants to the system and accessed them accordingly. Other possible values we could add are:
- `event`, `role`, `client`, `profile` (and all its sub keys), `variable_name` (specific to communication protocol), `safe_user_id`

### DIFF
--- a/learning_observer/learning_observer/constants.py
+++ b/learning_observer/learning_observer/constants.py
@@ -1,0 +1,9 @@
+'''
+Commonplace for learning observer constants.
+Many of these are dictionary keys we access frequently.
+Using the constants protect us from typos.
+'''
+AUTH_HEADERS = 'auth_headers'
+IMPERSONATING_AS = 'impersonating_as'
+USER = 'user'
+USER_ID = 'user_id'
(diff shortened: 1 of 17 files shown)


learning_observer/learning_observer/constants.py
```diff
@@ -0,0 +1,9 @@
+'''
+Commonplace for learning observer constants.
+Many of these are dictionary keys we access frequently.
+Using the constants protect us from typos.
```
pmitros, line 4: Add to comment: 

Over time, we plan to add migrate more constants here, but adding ALL constants here is not a goal. Many constants are better locally-scoped (or otherwise have good reason to live elsewhere).

pmitros:
Looks good to me! The constants file is pretty anaemic, but that's the right place to start. Everything in it clearly belongs in a constants file, and I'd rather add too little than too much on a first commit. A lot of other things (like `course_id`) should migrate there eventually, but we can get there on a case-by-case basis. 

I'd merge almost as-is:
* I'd added one line to the comment. An important strategy is to anticipate what idiots will do when they see your code, and then to guard against it in documentation. 
* It's also good to begin role-modelling documenting what constants are used for, and especially non-obvious ones.

PR #116, opened 2024-01-11
added addtional url support to regex to fix new doc not having an id bug

bradley-erickson:


pmitros:
Looks good to me. Feel free to merge. Or if you prefer, document the regexp somehow and then merge.

PR #117, opened 2024-01-11
added check for user in request while checking roles

bradley-erickson:


pmitros:
Looks good to me. Ready to merge.

PR #121, opened 2024-01-26
added rules overriding basic settings

bradley-erickson:


bradley-erickson:
This was the first attempt at refactoring some of the settings work. The settings work has moved passed this into its own module, PSS. Closing this PR

PR #126, opened 2024-03-06
Initial Fix for Log Files Issue

Brendon-Hablutzel:
# Initial Fix

Created an initial fix for the log file issue that works locally. I added a function `close_logfile` that closes log files and removes them from the system's list of open log files. This function is called when the websocket connection associated with those log files is closed.

For the files ending in `.study.log` (which are created in `event_decoder_and_logger`, this fix works well, as the log closing mechanism is simply included in the same function where the filename is generated.

However, for the files just ending in `.log` (which are created in `handle_incoming_client_event`), it is not ideal because it requires returning a filename from that function and closing the log file outside of the function (the function `handle_incoming_client_event` cannot determine whether the events have stopped being received, and therefore does not know when to close the log file).

This is because while `event_decoder_and_logger` returns a function that handles an iterable of events, and we can simply close the log file once there are no events remaining (e.g. the socket is closed), `handle_incoming_client_event` returns a handler that is called each time an event is received. This means that if we implemented something similar to with `event_decoder_and_logger` in `handle_incoming_client_event`, the log file would be closed after every single event.

I tried refactoring `handle_incoming_client_event` to be more similar to `event_decoder_and_logger`, returning a function that handles many events, rather than just one event, but this would require changing the code structure significantly.

Note: this addresses issue [#106](https://github.com/ArgLab/ArgLab_writing_observer/issues/106) on the ArgLab fork of the repo.

### DIFF
--- a/learning_observer/learning_observer/incoming_student_event.py
+++ b/learning_observer/learning_observer/incoming_student_event.py
@@ -423,12 +423,10 @@ async def incoming_websocket_handler(request):
         '''Prepare each event we receive for processing
         '''
         events = process_message_from_ws()
-        # NOTE: EVENT LOGS ARE USED HERE
         events = decoder_and_logger(events)
         events = decode_lock_fields(events)
         events = handle_auth_events(events)
         events = filter_blacklist_events(events)
-        # NOTE: STUDY LOGS ARE USED HERE
         events = pass_through_reducers(events)
         # empty loop to start the generator pipeline
         async for event in events:
--- a/learning_observer/learning_observer/incoming_student_event.py
+++ b/learning_observer/learning_observer/incoming_student_event.py
@@ -190,8 +190,7 @@ async def handle_incoming_client_event(metadata):
             json.dumps(event, sort_keys=True),
             filename, preencoded=True, timestamp=True)
         await pipeline(event)
-
-    return handler
+    return handler, filename
 
 
 COUNT = 0
@@ -282,6 +281,8 @@ def event_decoder_and_logger(
                 json_event = json.loads(msg.data)
             log_event.log_event(json_event, filename=filename)
             yield json_event
+        # done processing events, can close logfile now
+        log_event.close_logfile(filename)
     return decode_and_log_event
 
 
@@ -308,6 +309,7 @@ async def incoming_websocket_handler(request):
     lock_fields = {}
     authenticated = False
     event_handler = failing_event_handler
+    study_log_filename = None
 
     decoder_and_logger = event_decoder_and_logger(request)
 
@@ -335,13 +337,14 @@ async def incoming_websocket_handler(request):
             return
 
         nonlocal event_handler
+        nonlocal study_log_filename
         if 'source' in lock_fields:
             debug_log('Updating the event_handler()')
             metadata = lock_fields.copy()
         else:
             metadata = event
         metadata['auth'] = authenticated
-        event_handler = await handle_incoming_client_event(metadata=metadata)
+        event_handler, study_log_filename = await handle_incoming_client_event(metadata=metadata)
 
     async def handle_auth_events(events):
         '''This method checks a single method for auth and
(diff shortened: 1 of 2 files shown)


learning_observer/learning_observer/log_event.py
```diff
@@ -298,3 +298,7 @@ def log_ajax(url, resp_json, request):
     )
     with open(filename, "w") as ajax_log_fp:
         ajax_log_fp.write(encoded_payload)
+
+def close_logfile(filename):
+    # remove the file from the dict storing open log files and close it
+    files.pop(filename).close()
```
bradley-erickson, line 304: There should be some check that the filename exists in our open files before we attempt to close it. Not sure if we should throw an exception if it doesn't exist or just log a warning.

learning_observer/learning_observer/incoming_student_event.py
```diff
@@ -412,15 +415,20 @@ async def pass_through_reducers(events):
...
             yield event
+        # when the event handler is finished processing events, we can close the study log file
+        if study_log_filename is not None:
+            log_event.close_logfile(study_log_filename)
 
     async def process_ws_message_through_pipeline():
         '''Prepare each event we receive for processing
         '''
         events = process_message_from_ws()
+        # NOTE: EVENT LOGS ARE USED HERE
```
bradley-erickson, line 426: Remove these notes. They don't make much sense outside of the context of this pull request.

learning_observer/learning_observer/incoming_student_event.py
```diff
@@ -190,8 +190,7 @@ async def handler(request, client_event):
             json.dumps(event, sort_keys=True),
             filename, preencoded=True, timestamp=True)
         await pipeline(event)
-
-    return handler
+    return handler, filename
```
bradley-erickson, line 193: One downside of this approach is that we have this filename being passed around. The handler is also being passed around, but the fewer items like this the better (less to things to keep track of). Its better to keep all the event handler logic in one place rather than spread out.

One possibility for this is adding a subsequent function to the handler object that takes care of closing, so:
```python
event_handler = handle_incoming_client_event(...)
event_handler(request, event)  # normal behavior as it is now
event_handler.close() # new functionality that closes the file
```
I would be interest in @pmitros 's opinion on this implementation before diving head first into implementing this. What you currently have may be good enough for now.
To implement the functionality, feel free to take a gander at how the KVS is implemented on the system. You can call `kvs.KVS()` and get the default kvs or use `kvs.KVS.memoization()` to fetch a specific kvs.

Brendon-Hablutzel:
I made another commit fixing two of the issues:
- removed note comments
- add functionality for throwing an exception when the file is not found--piotr mentioned not using logging in the logging functionality itself, so I opted for an exception rather than a log message for now

learning_observer/learning_observer/incoming_student_event.py
pmitros: @bradley-erickson I'm okay with `.close()`, but I'm wondering if it's necessary. 

Have a look at: https://docs.python.org/3.6/library/weakref.html#weakref.finalize

learning_observer/learning_observer/log_event.py
```diff
@@ -298,3 +298,11 @@ def log_ajax(url, resp_json, request):
     )
     with open(filename, "w") as ajax_log_fp:
         ajax_log_fp.write(encoded_payload)
+
+def close_logfile(filename):
+    # remove the file from the dict storing open log files and close it
+    old_file = files.pop(filename)
+    if old_file is None:
+        raise KeyError(f"Tried to remove log file {old_file} but it was not found")
+    else:
```
bradley-erickson, line 307: The else statement here is unneeded. You can just do:
```python
if old_file is None:
    raise KeyError(...)
old_file.close()
```
If the error gets raised then the last line will not be ran, so need nesting it under an else.

PR #130, opened 2024-03-26
initial start of remaking makefile

bradley-erickson:
This removes extra npm commands we attempt to run and instead relies on the `Makefile` for installing the `LODRC`.


Makefile
```diff
@@ -1,65 +1,39 @@
...
-	#    pip install -r requirements.txt
-	#    cd learning_observer
-	#    python setup.py develop
-	#    python learning_observer
-	# And then clean up this Makefile to do it for you automatically :)
+install-dev: venv
+	# TODO create a dev requirements file
+	${PIP} install --no-cache-dir -e learning_observer/[${PACKAGES}]
+	. ${HOME}/.nvm/nvm.sh && nvm use && ${PIP} install -v -e modules/lo_dash_react_components/
+	${PIP} install ollama
```
bradley-erickson, line 33: - [ ] Need to add sphinx and all other items back in

modules/lo_dash_react_components/setup.py
```diff
@@ -1,16 +1,5 @@
 import json
 from setuptools import setup
```
bradley-erickson, line 2: We may want to store some of these commands when installing in dev mode

learning_observer/setup.py
```diff
@@ -25,48 +20,12 @@ def clean_requirements(filename):
...
     extras_require={
         "wo": clean_requirements(wo_path),
         "awe": clean_requirements(awe_path)
     },
-    cmdclass={
-        'develop': LOInstall,
-        'install': LOInstallTEST
-    }
+    packages=find_packages(),
+    package_data={'': ['util/*', 'static/**/*', 'static_data/*.template', 'creds.yaml.example', 'communication_protocol/schema.json']}
```
bradley-erickson, line 30: - [ ] Confirm there is nothing missing from here (this is needed for `pip install learning_observer/` without `-e` (dev mode).

Makefile
bradley-erickson: The prior makefile had an old way of handling sphinx

Makefile
```diff
@@ -1,65 +1,67 @@
...
-extension-package:
-	# I won't build this for you, but I will give you instructions, since
-	# the set of options will vary.
-	#
-	# On my last system, I needed:
-	#	google-chrome --pack-extension=extension --pack-extension-key=extension.pem --disable-setuid-sandbox --no-gpu --no-sandbox --headless
-	# On my current system
-	#	google-chrome --pack-extension=extension --pack-extension-key=extension.pem
-	# The pem file is your private key. You'll need to make one.
-	# The output will be called extension.crx
```
pmitros, line 25: Why is this removed? This should eventually be integrated into the `Makefile`, but until then, knowing how to build is useful.

Makefile
```diff
@@ -1,65 +1,67 @@
...
-	# can tolerate.
+venv:
+	virtualenv ${VENV}
+	${PIP} install --no-cache-dir -r requirements.txt
+clean-venv:
+	# Removing the virtual environment
+	rm -rf ${VENV}
 
-	pycodestyle --ignore=E501,W503 $(PYTHONFILES)
-	pylint -d W0613,W0511,C0301,R0913,too-few-public-methods $(PYTHONFILES)
```
pmitros, line 53: Why is this removed? It is helpful to be able to run `pycodestyle` and `pylint` with the right options.

Makefile
```diff
@@ -1,65 +1,67 @@
...
-	learning_observer/util/*py \
-	gitserve/gitserve/*py \
-	gitserve/*py \
-	learning_observer/learning_observer/pubsub/*py \
-	learning_observer/learning_observer/stream_analytics/*py \
-)
+PACKAGES ?= wo,awe
+VENV ?= .makevenv
+PYTHON = ${VENV}/bin/python
+PIP = ${VENV}/bin/pip
```
pmitros, line 4: I'm a little bit curious about the design decision to require the use of `virtualenv` and to manage it explicitly. 

In my version, I use `venv`, but maintain the abstraction barrier where the makefile doesn't know about it. Can you explain why this change is helpful?

Makefile
```diff
@@ -1,65 +1,67 @@
...
+	# Starting to lint Node modules
+	npm install
+	# Linting Javascript
+	npm run lint:js
+	# Linting CSS
+	npm run lint:css
+	# Finding any unused CSS files
+	npm run find-unused-css
+
+linting: linting-python
```
pmitros, line 60: Shouldn't this depend on linting-node as well?

Makefile
```diff
@@ -1,65 +1,67 @@
...
+# testing commands
+test:
+	# this is where we run doctests
+	pytest modules/wo_highlight_dashboard
+
+# Linting commands
+linting-python:
+	# Linting Python modules
+	${PIP} install pycodestyle
+	pycodestyle --ignore=E501,W503 $$(git ls-files 'learning_observer/*.py' 'modules/*.py')
```
pmitros, line 46: 1. It seems like having the `pip install` in the linting commands, rather than in system setup, would reduce performance. 
2. We seem to have lost the `pylint`
3. Why `git ls-files`? I'm not arguing -- just curious (versus a normal `find` or a normal wildcard)

pmitros:
On the whole, this looks good. I'd like to understand this better (questions left in-line; those aren't requests for changes, but I'd like to understand). I think the key changes I'd like to see:

1. Some way to build the extension (even the crude one of just printing out what to run)
2. `pylint` should be available as an option, with the right set of options
3. Either an understanding of why we manage `virtualenv` explicitly, or a move back to managing it implicitly. By my read, this adds complexity both to the code and to the use of the system (we can't leverage that people know how to use `virtualenv` themselves; they need to relearn our wrapper), and reduces flexibility (e.g. installing system-wide).

Makefile
bradley-erickson: This method of building is outdated. It seems as though the pack-extension commands are used for hosting the extension on its own server (instead of the Chrome store). The Chrome store expects a zip file to be uploaded. This is taken care of by `npm run build` in the extension directory. A command for that is at the bottom of the Makefile.

Makefile
bradley-erickson: I don't recall the exact steps that led me down preferring this path. I think it was helpful when testing. Relying on the user to create one and use it, or just use the overall environment, is probably the better workflow.

Makefile
bradley-erickson: 1. I'll add a linting-setup
2. I'll get it added back in - it did not make its way into the github action for linting so long ago and I don't recall exactly why.
3. Other commands were not playing well with github actions, `git ls-files` was the only command I could get to run in the github action environment. I copied some of that code over. Next steps after this include reworking some of the github actions to use makefiles.

pmitros: `pycodestyle` makes sense as a CI step since practically all issues it raises should be fixed. `pylint` shows a much longer list of _potential_ issues, which should be looked at, but many of which need not be addressed. 

Ergo, it's helpful to run manually sometimes in the `makefile`, but it should not be part of automation in the github action.

Makefile
pmitros: Makes sense. I wasn't aware this had changed.

PR #134, opened 2024-04-11
WiP: PMSS integration

bradley-erickson:
This pull request is a work in progress for integrating PMSS into the Learning Observer platform.

### DIFF
--- a/learning_observer/learning_observer/auth/social_sso.py
+++ b/learning_observer/learning_observer/auth/social_sso.py
@@ -41,6 +41,32 @@ import learning_observer.auth.roles
 import learning_observer.constants as constants
 import learning_observer.exceptions
 
+import pss
+pss.register_field(
+    name="hostname",
+    type=pss.psstypes.TYPES.hostname,
+    description="The hostname of the server",
+    required=True
+)
+pss.register_field(
+    name="protocol",
+    type=pss.psstypes.TYPES.protocol,
+    description="The protocol (http / https)",
+    required=True
+)
+pss.register_field(
+    name="client_id",
+    type=pss.psstypes.TYPES.string,
+    description="The Google OAuth client ID",
+    required=True
+)
+pss.register_field(
+    name="client_secret",
+    type=pss.psstypes.TYPES.string,
+    description="The Google OAuth client secret",
+    required=True
+)
+
 
 DEFAULT_GOOGLE_SCOPES = [
     'https://www.googleapis.com/auth/userinfo.profile',
@@ -59,6 +85,20 @@ DEFAULT_GOOGLE_SCOPES = [
     'https://www.googleapis.com/auth/documents.readonly',
     'https://www.googleapis.com/auth/classroom.announcements.readonly'
 ]
+
+# TODO Type list is not yet supported by PSS 4/11/24
+# pss.register_field(
+#     name='base_scopes',
+#     type='list',
+#     description='List of Google URLs to look for.',
+#     default=DEFAULT_GOOGLE_SCOPES
+# )
+# pss.register_field(
+#     name='additional_scopes',
+#     type='list',
+#     description='List of additional URLs to look for.',
+#     default=[]
+# )
 
 
 async def social_handler(request):
@@ -91,10 +131,10 @@ async def _google(request):
     if 'error' in request.query:
         return {}
 
-    hostname = settings.settings['hostname']
-    protocol = settings.settings.get('protocol', 'https')
+    hostname = settings.pss_settings.hostname()
+    protocol = settings.pss_settings.protocol()
     common_params = {
-        'client_id': settings.settings['auth']['google_oauth']['web']['client_id'],
+        'client_id': settings.pss_settings.client_id(types=['auth', 'google_oauth', 'web']),
         'redirect_uri': f"{protocol}://{hostname}/auth/login/google"
     }
 
@@ -125,7 +165,7 @@ async def _google(request):
     url = 'https://accounts.google.com/o/oauth2/token'
     params = common_params.copy()
     params.update({
-        'client_secret': settings.settings['auth']['google_oauth']['web']['client_secret'],
+        'client_secret': settings.pss_settings.client_secret(types=['auth', 'google_oauth', 'web']),
         'code': request.query['code'],
         'grant_type': 'authorization_code',
     })
--- a/learning_observer/learning_observer/auth/social_sso.py
+++ b/learning_observer/learning_observer/auth/social_sso.py
@@ -42,16 +42,19 @@ import learning_observer.constants as constants
 import learning_observer.exceptions
 
 import pss
+# TODO the hostname setting currently expect the port
+# to specified within the hostname. We ought to
+# remove the port and instead use the port setting.
 pss.register_field(
     name="hostname",
     type=pss.psstypes.TYPES.hostname,
-    description="The hostname of the server",
+    description="The hostname of the LO webapp. Used to redirect OAuth clients.",
     required=True
 )
 pss.register_field(
     name="protocol",
     type=pss.psstypes.TYPES.protocol,
-    description="The protocol (http / https)",
+    description="The protocol (http / https) of the LO webapp. Used to redirect OAuth clients.",
     required=True
 )
 pss.register_field(
--- a/learning_observer/learning_observer/auth/social_sso.py
+++ b/learning_observer/learning_observer/auth/social_sso.py
@@ -41,31 +41,31 @@ import learning_observer.auth.roles
 import learning_observer.constants as constants
 import learning_observer.exceptions
 
-import pss
+import pmss
 # TODO the hostname setting currently expect the port
 # to specified within the hostname. We ought to
 # remove the port and instead use the port setting.
-pss.register_field(
+pmss.register_field(
     name="hostname",
-    type=pss.psstypes.TYPES.hostname,
+    type=pmss.pmsstypes.TYPES.hostname,
     description="The hostname of the LO webapp. Used to redirect OAuth clients.",
     required=True
 )
-pss.register_field(
+pmss.register_field(
     name="protocol",
-    type=pss.psstypes.TYPES.protocol,
+    type=pmss.pmsstypes.TYPES.protocol,
     description="The protocol (http / https) of the LO webapp. Used to redirect OAuth clients.",
     required=True
 )
-pss.register_field(
+pmss.register_field(
     name="client_id",
-    type=pss.psstypes.TYPES.string,
+    type=pmss.pmsstypes.TYPES.string,
     description="The Google OAuth client ID",
     required=True
 )
-pss.register_field(
+pmss.register_field(
     name="client_secret",
-    type=pss.psstypes.TYPES.string,
+    type=pmss.pmsstypes.TYPES.string,
     description="The Google OAuth client secret",
     required=True
 )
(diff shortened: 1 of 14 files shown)


modules/writing_observer/writing_observer/aggregator.py
```diff
@@ -339,6 +352,10 @@ async def fetch_doc_from_google(student):
     return writing_data
 
 
+# NOTE This is old way of querying data from the system.
+# The code should all still function, but the proper way to
+# do this is using the Communication Protocol.
+# This function and any references should be removed.
```
bradley-erickson, line 358: Any comments?

modules/writing_observer/writing_observer/aggregator.py
```diff
@@ -218,7 +231,7 @@ async def merge_with_student_data(writing_data, student_data):
     return writing_data
 
 
-use_nlp = learning_observer.settings.module_setting('writing_observer', 'use_nlp', False)
+use_nlp = learning_observer.settings.pss_settings.use_nlp(types=['modules', 'writing_observer'])
```
bradley-erickson, line 234: Using this at the top level of the file seems out of place and may break during future validation steps. This and subsequent lines ought to live in a function that's triggered during startup.

bradley-erickson:
Some settings in the system do a lot of checking to see if an object exists in the settings.
Code similar to `"google_oauth" in learning_observer.settings.settings['auth']` appears in a few different places and its not clear how we should handle that within LO or PSS. `google_oauth` is a key along a nested dictionary here.

A similar issue is handling dictionaries in general. Some pieces on LO expect a dictionary coming back from settings. We can easily change something like

```python
sub_dict = settings.settings['item_that_returns_dict']
# TO
sub_dict = {
    'val1': settings.settings.val1(types=['item_that_returns_dict']),
    # ... the rest of the dictionary items
}
```

The most complex use case for settings I've encountered thus far is with the KVS. We have

```python
KVS = KVSRouter(items=learning_observer.settings.settings['kvs'].items())
```

with settings that look like:

```yaml
kvs:
  default:
    type: redis_ephemeral
    expiry: 86400
  memoization:
    type: redis_ephemeral
    expiry: 60
  ...
```

I'm not sure the best way to handle this use case.

modules/writing_observer/writing_observer/aggregator.py
bradley-erickson: This is not specific to the PSS integration, but something I noticed while swapping some settings over to PSS. This is how we fetched `latest_data()` before the communication protocol existed.

bradley-erickson:
Feature flags in the system seem a little messy at the moment.
Some pieces of the code want to check if a certain item is available or not in the dictionary of feature flags. For fetching a specific setting, the code uses `settings.feature_flag(flag)`. This code attempts to fetch the appropriate feature flag value.

Under the PSS integration, should we 
1) modify the `feature_flag` method, so we can continue calling
    `flag_setting = settings.feature_flag(flag)` OR
2) manually call the items with pss
    `flag_setting = settings.pss_settings.flag(types=['feature_flag'])`

I think 1 is the correct option here. The modules should probably follow a similar route. At the time of writing this, I already changed the modules to use the 2 workflow (without initially thinking about 1).

learning_observer/learning_observer/auth/social_sso.py
```diff
@@ -91,10 +131,10 @@ async def _google(request):
     if 'error' in request.query:
         return {}
 
-    hostname = settings.settings['hostname']
-    protocol = settings.settings.get('protocol', 'https')
+    hostname = settings.pss_settings.hostname()
+    protocol = settings.pss_settings.protocol()
```
pmitros, line 135: I think we'll eventually want short syntax. e.g. `pss.hostname()` or `settings.hostname()` or similar. 

Doesn't need to be this PR.

learning_observer/learning_observer/auth/social_sso.py
```diff
@@ -91,10 +131,10 @@ async def _google(request):
...
     common_params = {
-        'client_id': settings.settings['auth']['google_oauth']['web']['client_id'],
+        'client_id': settings.pss_settings.client_id(types=['auth', 'google_oauth', 'web']),
```
pmitros, line 137: Reading this, I hate the word `types`. It makes sense, since it comes from css types. We should brainstorm about better ways to do this. Again, doesn't need to be this PR.

learning_observer/learning_observer/dashboard.py
```diff
@@ -35,6 +36,13 @@
 import learning_observer.communication_protocol.schema
 import learning_observer.settings
 
+pss.register_field(
+    name='dangerously_allow_insecure_dags',
+    type=pss.psstypes.TYPES.boolean,
+    description='Determine whether we should allow self-made dags on the system.',
```
pmitros, line 42: Doesn't need to be this PR, but we will want clearer descriptions. A new user won't know what a self-made DAG is.

learning_observer/learning_observer/main.py
```diff
@@ -69,10 +81,9 @@ def create_app():
     # We don't want these to change on a restart.
     # We should check if reloading this module overwrites them.
     if port is None:
-        port = settings.settings.get("server", {}).get("port", None)
+        port = settings.pss_settings.port(types=['server'])
```
pmitros, line 84: `server` is a little generic. Perhaps `lo_webapp` or similar?

learning_observer/learning_observer/pubsub/__init__.py
```diff
@@ -15,14 +15,26 @@
...
 https://github.com/encode/broadcaster
 '''
-
+import pss
 import sys
 
 import learning_observer.settings as settings
 from learning_observer.log_event import debug_log
 
+@pss.parser('pubsub_type', parent='string', choices=['xmpp', 'stub', 'redis'])
```
pmitros, line 24: We should probably just kill pubsub, since we don't use it anymore.

learning_observer/learning_observer/pubsub/__init__.py
```diff
@@ -15,14 +15,26 @@
...
+pss.register_field(
+    name='type',
+    type='pubsub_type',
+    description='The type of pub-sub model to use on the system.',
+    default='stub'
+)
+
 try:
     PUBSUB = settings.settings['pubsub']['type']
+    PUBSUB = settings.pss_setting.type(types=['pubsub'])
```
pmitros, line 37: `type` is a very poor name here. We can change it in a future PR or right now.

learning_observer/learning_observer/rosters.py
```diff
@@ -85,6 +86,17 @@
 COURSE_URL = 'https://classroom.googleapis.com/v1/courses'
 ROSTER_URL = 'https://classroom.googleapis.com/v1/courses/{courseid}/students'
 
+@pss.parser('roster_source', parent='string', choices=['google_api', 'all', 'test', 'filesystem'])
+def _convert_roster_source(value):
+    return value
```
pmitros, line 91: This is a little bit ugly. I'm wondering if it might make sense to define an `identity` function and have something like:

`pss.parser('roster_source', parent='string', choices=['google_api', 'all', 'test', 'filesystem'])(identity)`

Or perhaps give `pss.parser` some way to not be used as an operator e.g.

`pss.parser('roster_source', parent='string', choices=['google_api', 'all', 'test', 'filesystem'], transform=None)`

or

`pss.parser('roster_source', parent='string', choices=['google_api', 'all', 'test', 'filesystem'])(None)`

or similar.

learning_observer/learning_observer/rosters.py
```diff
@@ -85,6 +86,17 @@
...
+
+pss.register_field(
+    name='source',
+    type='roster_source',
+    description='Source to use for rosters.',
```
pmitros, line 96: At the very least, "student class rosters" or similar

modules/writing_observer/writing_observer/aggregator.py
pmitros: That's fair. This should all probably happen in a registered initialization function. I might leave a comment with a TODO, but address in a future PR.

pmitros:
> Feature flags in the system seem a little messy at the moment. Some pieces of the code want to check if a certain item is available or not in the dictionary of feature flags. For fetching a specific setting, the code uses `settings.feature_flag(flag)`. This code attempts to fetch the appropriate feature flag value.
> 
> Under the PSS integration, should we
> 
>     1. modify the `feature_flag` method, so we can continue calling
>        `flag_setting = settings.feature_flag(flag)` OR
> 
>     2. manually call the items with pss
>        `flag_setting = settings.pss_settings.flag(types=['feature_flag'])`
> 
> 
> I think 1 is the correct option here. The modules should probably follow a similar route. At the time of writing this, I already changed the modules to use the 2 workflow (without initially thinking about 1).

I would agree that we should stay with #1. There could be a thin wrapper around however we handle this in `pss`, and eventually, this might make it into `pss` as a first-class sort of thing (which might go together with lists and dictionaries more broadly).

pmitros:
> Some settings in the system do a lot of checking to see if an object exists in the settings. Code similar to `"google_oauth" in learning_observer.settings.settings['auth']` appears in a few different places and its not clear how we should handle that within LO or PSS. `google_oauth` is a key along a nested dictionary here.
> 
> A similar issue is handling dictionaries in general. Some pieces on LO expect a dictionary coming back from settings. We can easily change something like
> 
> ```python
> # Currently looks like:
> sub_dict = settings.settings['item_that_returns_dict']
> 
> # Could look like the following instead
> sub_dict = {
>     'val1': settings.settings.val1(types=['item_that_returns_dict']),
>     # ... the rest of the dictionary items
> }
> # we build the dictionary in the code when we read in the settings
> ```
> 
> The most complex use case for settings I've encountered thus far is with the KVS. We have
> 
> ```python
> KVS = KVSRouter(items=learning_observer.settings.settings['kvs'].items())
> ```
> 
> with settings that look like:
> 
> ```yaml
> kvs:
>   default:
>     type: redis_ephemeral
>     expiry: 86400
>   memoization:
>     type: redis_ephemeral
>     expiry: 60
>   ...
> ```
> 
> I'm not sure the best way to handle this use case. For now I'll leave it as is.
> 
> EDIT: The `event_auth` goes through a similar process of initializing itself.
> 
> ```python
>     for auth_method in learning_observer.settings.settings['event_auth']:
> ```

We had walked through several ways this might be handled, but IIRC, I think the strategy we converged on was to leave as-is, with list and dictionary-style settings continuing to live in `creds.yaml`, and moving over incrementally things which fit into `pss` to use `pss`, and finishing the complex pieces in a further PR.

modules/writing_observer/writing_observer/aggregator.py
pmitros: I'd convert this from a NOTE to a TODO and leave for later :)

pmitros:
Notes:

* We could have a way to iterate for every type in a selector with some other type.
* We could have a way to iterate through children. Currently, `pss` treats `auth google` and `google auth` as identical. We could not do that, although we'd lost some niceness.
* We could use the direct child selector `auth>google` (although we lose the ability to swap order)
* We could use the sibling selector `auth+google` (where we maintain the ability)
* We could be explicit, and have a list of enabled auth methods. It may make sense to have configurations for ones we don't use.
* We could introduce JSON-style list types: ` auth { methods: ["google","roster","test"]; }`
* We could introduce a simpler list style of some kind
* We could introduce accumulating types, where `auth { methods: google; } auth { methods: roster; } ` results in the selector `auth methods` returning both `google` and `roster`. In css, this would override. Accumulation would be an expansion. 

We could do more than one of the above too.

bradley-erickson:
- [ ] PSS needs to be published in some fashion before merging

PR #136, opened 2024-04-25
Arglab reconstruction fixes

bradley-erickson:
These are various fixes courtesy of @Code-beep12 to the reconstruction workflows.

This ought to get merged in soon. These commits are cherry-picked from the original branch. The original branch included some merge commits we did not want to include in this pull request.

### DIFF
--- a/modules/writing_observer/writing_observer/reconstruct_doc.py
+++ b/modules/writing_observer/writing_observer/reconstruct_doc.py
@@ -20,7 +20,6 @@ class google_text(object):
     point in time. Right now, this adds cursor position. In the future,
     we might annotate formatting and similar properties.
     '''
-
     def __new__(cls):
         '''
         Constructor. We create a blank document to be populated.
@@ -42,7 +41,9 @@ class google_text(object):
         textlength_array_length = len(self._edit_metadata["length"])
         length_difference = cursor_array_length - textlength_array_length
         if length_difference != 0:
-            raise Exception("Edit metadata length doesn't match. This should never happen.")
+            raise Exception(
+                "Edit metadata length doesn't match. This should never happen."
+            )
 
     def fix_validity(self):
         '''
@@ -145,7 +146,11 @@ class google_text(object):
         '''
         This serializes to JSON.
         '''
-        return {'text': self._text, 'position': self._position, 'edit_metadata': self._edit_metadata}
+        return {
+            'text': self._text,
+            'position': self._position,
+            'edit_metadata': self._edit_metadata
+        }
 
     def get_parsed_text(self):
         '''
@@ -211,7 +216,11 @@ def insert(doc, ty, ibi, s):
     if ibi > nextchar_index:
         insert(doc, ty, nextchar_index, PLACEHOLDER * (ibi - nextchar_index))
 
-    doc.update("{start}{insert}{end}".format(start=doc._text[0 : ibi - 1], insert=s, end=doc._text[ibi - 1 :]))
+    doc.update("{start}{insert}{end}".format(
+        start=doc._text[0:ibi - 1],
+        insert=s,
+        end=doc._text[ibi - 1:]
+    ))
 
     doc.position = ibi + len(s)
 
@@ -236,7 +245,10 @@ def delete(doc, ty, si, ei):
     if ei > lastchar_index:
         insert(doc, ty, lastchar_index + 1, PLACEHOLDER * (ei - lastchar_index))
 
-    doc.update("{start}{end}".format(start=doc._text[0 : si - 1], end=doc._text[ei:]))
+    doc.update("{start}{end}".format(
+        start=doc._text[0:si - 1],
+        end=doc._text[ei:]
+    ))
 
     doc.position = si
 
--- a/modules/writing_observer/writing_observer/reconstruct_doc.py
+++ b/modules/writing_observer/writing_observer/reconstruct_doc.py
@@ -64,9 +64,6 @@ class google_text(object):
         '''
         errors_found = []
 
-        # store image indexes
-        self._edit_metadata['images'] = {}
-
         if "cursor" not in self._edit_metadata:
             self._edit_metadata["cursor"] = []
             errors_found.append("No cursor array")
@@ -166,26 +163,9 @@ class google_text(object):
 
     def get_parsed_text(self):
         '''
-        Returns the text ignoring the image placeholders as well as
-        normal placeholders
+        Returns the text ignoring the normal placeholders
         '''
-        parsed_text = self._text.replace(PLACEHOLDER, "") #remove all placeholder characters
-        new_text = []
-        #ignore all the image indexes
-        for idx, s in enumerate(parsed_text, start=1):
-            if idx in self._edit_metadata['images'].values():
-                continue
-            new_text.append(s)
-        return ''.join(new_text)
-
-    def update_image_index(self, si, offset):
-        '''
-        Updates the image index by offset characters
-        Called by insert() and delete() events
-        '''
-        for image_id, idx in self._edit_metadata['images'].items():
-            if si <= idx:
-                self._edit_metadata['images'][image_id] += offset
+        return self._text.replace(PLACEHOLDER, "")
 
 
 def command_list(doc, commands):
@@ -236,8 +216,6 @@ def insert(doc, ty, ibi, s):
 
     doc.position = ibi + len(s)
 
-    doc.update_image_index(ibi, len(s))
-
     return doc
 
 
@@ -264,9 +242,6 @@ def delete(doc, ty, si, ei):
 
     doc.position = si
 
-    offset = ei - si + 1
-    doc.update_image_index(si, -offset)
-
     return doc
 
 
@@ -276,31 +251,6 @@ def alter(doc, si, ei, st, sm, ty):
 
     We ignore these for now.
     '''
-    return doc
-
-
-def image_index(doc, ty, id, spi):
-    '''
-    Called whenever an image is added or when an image's position is changed
-    * `ty` is always `te`
-    * `id` is the unique image id
-    * `spi` is the image index
-    '''
-    doc.edit_metadata['images'][id] = spi
-    return doc
-
-
-def image_delete(doc, ty, id, et):
-    '''
-    Called whenever an image is deleted
-    * `ty` is always `de`
-    * `id` is the unique image id
-    '''
-    try:
-        doc.edit_metadata['images'].pop(id)
-    except KeyError:
-        # This happens when logfile is incomplete
-        pass
     return doc
 
 
--- a/modules/writing_observer/writing_observer/reconstruct_doc.py
+++ b/modules/writing_observer/writing_observer/reconstruct_doc.py
@@ -9,6 +9,7 @@ See: `http://features.jsomers.net/how-i-reverse-engineered-google-docs/`
 
 import json
 
+PLACEHOLDER = '\x00'
 
 class google_text(object):
     '''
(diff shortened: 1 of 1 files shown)


modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -37,9 +42,7 @@ def assert_validity(self):
         textlength_array_length = len(self._edit_metadata["length"])
         length_difference = cursor_array_length - textlength_array_length
         if length_difference != 0:
-            raise Exception(
-                "Edit metadata length doesn't match. This should never happen."
-            )
+            raise Exception("Edit metadata length doesn't match. This should never happen.")
```
bradley-erickson, line 45: This and other formatting should remain as it was.
This specific instance is less important.

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -179,14 +204,19 @@ def insert(doc, ty, ibi, s):
...
-        end=doc._text[ibi - 1:]
-    ))
+    # The index of the next character after the last character of the text
+    nextchar_index = len(doc._text) + 1
+    # If the insert index is greater than nextchar_index, insert placeholders to fill the gap
+    # This occurs when the document has undergone modifications before the logger has been initialized
+    if ibi > nextchar_index:
+        insert(doc, ty, nextchar_index, PLACEHOLDER * (ibi - nextchar_index))
+
+    doc.update("{start}{insert}{end}".format(start=doc._text[0 : ibi - 1], insert=s, end=doc._text[ibi - 1 :]))
```
bradley-erickson, line 214: The prior format of this code is more readable. Did a linter do this?

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -9,13 +9,18 @@
 
 import json
 
+# The placeholder character that is used for filling gaps in the document.
+# (occurs when there is mismatch between the index and the doc._text length)
+PLACEHOLDER = '\x00'
```
bradley-erickson, line 14: This needs a bit more explanation in the code as to what it is and what we need to do with it. A written out example in the comments would be nice.

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -9,13 +9,18 @@
...
+
 
 class google_text(object):
     '''
     We encapsulate a string object to support a Google Doc snapshot at a
     point in time. Right now, this adds cursor position. In the future,
     we might annotate formatting and similar properties.
     '''
+
```
bradley-erickson, line 23: If the linter complained about this, it's fine. Try to avoid adding in random new lines throughout the codebase unless a linter is complaining.

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -139,11 +145,30 @@ def json(self):
...
+        parsed_text = self._text.replace(PLACEHOLDER, "") #remove all placeholder characters
+        new_text = []
+        #ignore all the image indexes
+        for idx, s in enumerate(parsed_text, start=1):
+            if idx in self._edit_metadata['images'].values():
+                continue
+            new_text.append(s)
+        return ''.join(new_text)
+
+    def update_image_index(self, si, offset):
```
bradley-erickson, line 164: What do the images objects look like? Is it possible to record the image source?

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -139,11 +145,30 @@ def json(self):
...
         This serializes to JSON.
         '''
-        return {
-            'text': self._text,
-            'position': self._position,
-            'edit_metadata': self._edit_metadata
-        }
+        return {'text': self._text, 'position': self._position, 'edit_metadata': self._edit_metadata}
+
+    def get_parsed_text(self):
```
bradley-erickson, line 150: This method will make more sense to me once the placeholder example is added above.

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -197,13 +227,22 @@ def delete(doc, ty, si, ei):
...
+    #Index of the last character in the text. `si` and `ei` shouldn't go beyond that
+    lastchar_index = len(doc._text)
+    # If the deletion indexes are greater than nextchar_index, insert placeholders to fill the gap
+    # This occurs when the document has undergone modifications before the logger has been initialized
+    if si > lastchar_index:
+        insert(doc, ty, lastchar_index + 1, PLACEHOLDER * (si - lastchar_index))
+    if ei > lastchar_index:
+        insert(doc, ty, lastchar_index + 1, PLACEHOLDER * (ei - lastchar_index))
+
+    doc.update("{start}{end}".format(start=doc._text[0 : si - 1], end=doc._text[ei:]))
```
bradley-erickson, line 239: This looks awfully similar to some of the code above it. Should this be moved to its function? What is different about them?

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -197,13 +227,22 @@ def delete(doc, ty, si, ei):
...
 
     doc.position = si
 
+    offset = ei - si + 1
+    doc.update_image_index(si, -offset)
```
bradley-erickson, line 244: What is this offset for?

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -216,6 +255,31 @@ def alter(doc, si, ei, st, sm, ty):
...
+def image_delete(doc, ty, id, et):
+    '''
+    Called whenever an image is deleted
+    * `ty` is always `de`
+    * `id` is the unique image id
+    '''
+    try:
+        doc.edit_metadata['images'].pop(id)
+    except KeyError:
+        # This happens when logfile is incomplete
```
bradley-erickson, line 278: I'm confused by this comment. This code seems like it should `KeyError` when the image is not present in `doc.edit_metadata['images']`, but the comment states it has to do with a logfile. What is the logfile referring to here?

Code-beep12: `logfile` represents the logs. KeyError is raised when image is not present in the keys. This happens when there is no corresponding `Image Add` event (`ae`) for an image ID, but there is an `Image Delete` event (`de`) on that ID in the logs. 
For example, say there is a google doc that contains an image that was added before the writing_observer extension was started. The logs now lack the `image add` event on this image.
When a user later removes the image, it'll cause a KeyError as the logs have no record of this image.

modules/writing_observer/writing_observer/reconstruct_doc.py
Code-beep12: Yes my linter did it. I'll format it according to the previous format used in the project

modules/writing_observer/writing_observer/reconstruct_doc.py
Code-beep12: Each image stored in  self._edit_metadata['images'] has the image id and index of the image

modules/writing_observer/writing_observer/reconstruct_doc.py
Code-beep12: This basically adjusts the image indexes stored internally in `self._edit_metadata['images']` to sync with the document.
When a character is deleted, all the indexes after it will be offset by -1. This offset here is used to update the image index if necessary

modules/writing_observer/writing_observer/reconstruct_doc.py
bradley-erickson: 1. What is the image id used for? i.e. see if there are any Google endpoints 
2. Is there any information we are skipping from the event stream? If there is, we should record it.

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -179,6 +201,13 @@ def insert(doc, ty, ibi, s):
     * `ibi` is where the insert happens
     * `s` is the string to insert
     '''
+    # The index of the next character after the last character of the text
+    nextchar_index = len(doc._text) + 1
+    # If the insert index is greater than nextchar_index, insert placeholders to fill the gap
+    # This occurs when the document has undergone modifications before the logger has been initialized
+    if ibi > nextchar_index:
+        insert(doc, ty, nextchar_index, PLACEHOLDER * (ibi - nextchar_index))
```
bradley-erickson, line 209: More comments on why we are doing some as opposed to what the code is doing. Reading the code tells us what the code is doing, comments tell us why we are doing it.

bradley-erickson: Providing some context as to what the code does is good, but comments are better for the why.

Apply this to all comments in the pull request (not just the ones this comment is under).

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -197,6 +226,15 @@ def delete(doc, ty, si, ei):
...
     * `ei` is the end
     '''
+    #Index of the last character in the text. `si` and `ei` shouldn't go beyond that
+    lastchar_index = len(doc._text)
+    # If the deletion indexes are greater than nextchar_index, insert placeholders to fill the gap
+    # This occurs when the document has undergone modifications before the logger has been initialized
+    if si > lastchar_index:
+        insert(doc, ty, lastchar_index + 1, PLACEHOLDER * (si - lastchar_index))
+    if ei > lastchar_index:
+        insert(doc, ty, lastchar_index + 1, PLACEHOLDER * (ei - lastchar_index))
```
bradley-erickson, line 236: I thought I made a comment about this earlier, but I no longer see it. These if statements look awfully similar to the if statement used above. Should these 2 lines of code that are repeated 3 times be abstracted up a level to their own method?

bradley-erickson:
I also found some bugs in the placeholder text workflow. Here are the steps I took to break it:

1. Start typing in the document (call this text A)
1. Turn off the extension
1. Start typing earlier in the document (before text A)
1. Turn on extension
1. Start typing in the middle of text A (call this text B).
1. Examine reconstruction text in redis

The reconstruction text will be text A followed by some placeholders and then Text B will appear even though we placed text B in the middle of text A.

Code-beep12:
Unless the text you added before text A is present in the logs, there's no way of knowing the right index of text B.

Say you type in "alpha". Then you type "beta" before "alpha" when the extension is off.
Now when you type in something between "alpha", say after the third character. The index of the new text becomes 4+4=8 (length of "beta" + 4). Since "beta" is not recorded in the logs, the new text will be seen as being added after "alpha".

modules/writing_observer/writing_observer/reconstruct_doc.py
Code-beep12: Sure, I'll add them to its own function

bradley-erickson:
That workflow highlights a way that the reconstruction would break. This workflow and any others ought to have some thought put into how we can combat the breaking behavior.

I don't think this is something worth fixing in this pull request, just something to think about as you continue working.

modules/writing_observer/writing_observer/reconstruct_doc.py
```diff
@@ -197,6 +226,15 @@ def delete(doc, ty, si, ei):
     * `si` is the index of the start of deletion
     * `ei` is the end
     '''
+    #Index of the last character in the text. `si` and `ei` shouldn't go beyond that
```
pmitros, line 229: Nit: Space after #

PR #137, opened 2024-04-28
WiP: Update to lo_event to subscribe to have functions to help subscribe to many events.

pmitros:
Currently, we want to log events like typing, cut-and-paste, and debug information. This is a start of a PR to have a comprehensive logging framework for most browser events.


.gitignore
```diff
@@ -22,10 +22,19 @@ learning_observer/learning_observer/static_data/repos/
...
 .venv/
 .vscode/
 build/
 dist/
 node_modules
+*.orig
+lo_event*tgz
+*.log
+LanguageTool-stable.zip
+LanguageTool-5.4
```
bradley-erickson, line 36: Do these need to be in the main gitignore? They feel specific to your workstation.

modules/lo_event/lo_event/prototype/util.js
```diff
@@ -0,0 +1,77 @@
...
+     "hello.bar"
+  )
+
+  This can also handle embbedded lists identified
+  using notations like addedNodes[0].className.
+
+  If not found, return null
+
+  This was created in the extension, but should be merged
+  into `lo_event`
```
bradley-erickson, line 15: This should be marked as a TODO or handled in this PR

modules/lo_event/lo_event/prototype/events.js
```diff
@@ -0,0 +1,294 @@
...
+  logToDiv(`<b>${eventType}</b>: `+JSON.stringify(lodict));
+}
+
+export function subscribeToEvents({ target = document, eventList = events } = {}) {
+  for (let key in eventList) {
+    target.addEventListener(key, eventListener);
+  }
+}
+
+subscribeToEvents();
```
bradley-erickson, line 294: what is the difference between `browser_events.js` and `events.js`?

.gitignore
pmitros: 1. Password files should always be in gitignore. That's the format we use. 
2. *.orig, I have mixed feelings about. That's how git handles merge conflicts. These are sometimes helpful in `git status`, but should not be committed by accident.
3. lo_event*tgz is our build tarball for installing lo_event
4. *.log are the log files we generate
5. LanguageTool should probably not be version-locked, but we often download and run it to use some of the dashboards.  
6. package-lock.json is in our NPM packages. I generally avoid committing these, but it's a possible point-of-discussion.
7. static_data is where we keep snapshots from Google AJAX for local replay, as well as lists of teachers, admins, etc. We don't want to commit any of that.

modules/lo_event/lo_event/prototype/events.js
pmitros: Bad commit. This should have been a rename from events to browser_events. Coincidentally, most of the review should be on code there-in.

modules/lo_event/lo_event/prototype/util.js
pmitros: Agreed. It's a work-in-=progress. :)

PR #138, opened 2024-04-30
added binned time on task

bradley-erickson:


bradley-erickson:
break up to named functions, so things are more readability.

modules/writing_observer/writing_observer/writing_analysis.py
```diff
@@ -35,6 +35,8 @@
 # Should be 60-300 in prod. 5 seconds is nice for debugging
 TIME_ON_TASK_THRESHOLD = 5
 
+TIME_ON_TASK_BIN_SIZE = 600 # 10 minutes
```
pmitros, line 38: Let's grab these from settings, with some default.

modules/writing_observer/writing_observer/writing_analysis.py
```diff
@@ -94,6 +96,54 @@ async def time_on_task(event, internal_state):
...
+    Similar to the `time_on_task` reducer defined above, except it
+    bins the time spent.
+    '''
+    if internal_state is None:
+        internal_state = {
+            'saved_ts': None,
+            'binned_time_on_task': {},
+            'current_bin': None
+        }
+    last_ts = internal_state['saved_ts']
```
pmitros, line 111: I usually write code like this when I'm coding, and when I'm done, I do a search-and-replace with full variable names like `current_bin` rather than `curr_bin` before merge.

bradley-erickson: `timestamp` instead of `ts`

modules/writing_observer/writing_observer/writing_analysis.py
```diff
@@ -94,6 +96,54 @@ async def time_on_task(event, internal_state):
...
+    curr_bin = internal_state['current_bin']
+    internal_state['saved_ts'] = event['server']['time']
+
+    # Initial conditions
+    if last_ts is None:
+        last_ts = internal_state['saved_ts']
+    if curr_bin is None:
+        curr_bin = int((last_ts // TIME_ON_TASK_BIN_SIZE) * TIME_ON_TASK_BIN_SIZE)
```
pmitros, line 119: This code is profoundly unclear. I would have a function, along the lines of `bin(timestamp)` or `bin_index(timestamp)`

The combination of `int` and `//` feels confused (if not necessarily incorrect).

modules/writing_observer/writing_observer/writing_analysis.py
```diff
@@ -94,6 +96,54 @@ async def time_on_task(event, internal_state):
...
+    next_bin = curr_bin + TIME_ON_TASK_BIN_SIZE
+    next_bin_str = str(next_bin)
```
pmitros, line 121: `str(int(next_bin))` or possible a format string of some kind, rather than in the computation?

modules/writing_observer/writing_observer/writing_analysis.py
```diff
@@ -94,6 +96,54 @@ async def time_on_task(event, internal_state):
...
+    if last_ts is not None:
+        delta_t = min(
+            TIME_ON_TASK_THRESHOLD,               # Maximum time step
```
pmitros, line 124: pmss

pmitros:
There are a lot of clean-ups which could happen, but I don't think this code matters very much. So long as the interface is good, we're good. I might:

1) Move constants into `pmss`, if not too hard

2) Provide better variable names

3) Move `int` one online down, into the formatting line

4) If time allows, break things out into named functions

Merge. If time doesn't allow, we can skip any of steps 1-4, and just add a comment that this is write-only-code (in more words than that).

PR #140, opened 2024-05-02
WiP: initial doc processing file

bradley-erickson:


learning_observer/learning_observer/auth/social_sso.py
```diff
@@ -78,12 +82,41 @@ async def social_handler(request):
...
+    '''
+    kvs = learning_observer.kvs.KVS()
+    # store teacher auth info
+    auth_key = sa_helpers.make_key(
+        learning_observer.auth.utils.google_stored_auth,
+        {sa_helpers.KeyField.TEACHER: id},
+        sa_helpers.KeyStateType.INTERNAL)
+    await kvs.set(auth_key, request[constants.AUTH_HEADERS])
+
+    # store teacher roster info
```
pmitros, line 107: I'm curious why roster is stored and passed, rather than retrieved in the background process. It feels like the whole compute DAG should happen there.

learning_observer/learning_observer/auth/utils.py
```diff
@@ -99,13 +101,28 @@ async def get_active_user(request):
     return session[constants.USER]
 
 
+def google_stored_auth():
```
pmitros, line 104: I might consider using  `_store_teacher_info_for_background_process` (perhaps renamed).

learning_observer/learning_observer/auth/utils.py
```diff
@@ -99,13 +101,28 @@ async def get_active_user(request):
...
     session.pop(constants.AUTH_HEADERS, None)
+
+    # Remove auth header info from redis
+    kvs = learning_observer.kvs.KVS()
+    remove_auth_key = sa_helpers.make_key(
+        google_stored_auth,
+        {sa_helpers.KeyField.TEACHER: user['user_id']},
+        sa_helpers.KeyStateType.INTERNAL)
+    await kvs.remove(remove_auth_key)
+
```
pmitros, line 125: I would probably remove this when auth expires (e.g. we make a request and it fails with an auth exception), rather than on logout.

learning_observer/learning_observer/doc_processor.py
```diff
@@ -0,0 +1,235 @@
...
+- [ ] teacher visited dashboard recently
+- [ ] events > 2*[events since last processed]
+Start with just the first one
+'''
+async def check_recent_mod_and_not_recent_process(doc_id):
+    '''Determine if the document has been altered since its last
+    processing and check whether it is past a specified cutoff
+    time (5 minutes).
+    '''
+    cutoff = 300 # 5 minutes
```
pmitros, line 37: I'd take this from settings.

learning_observer/learning_observer/doc_processor.py
```diff
@@ -0,0 +1,235 @@
...
+        sa_helpers.KeyStateType.INTERNAL
+    )
+    doc_info = await KVS[key]
+    if doc_info is None: return True
+
+    key = sa_helpers.make_key(
+        writing_observer.writing_analysis.time_on_task,
+        {sa_helpers.EventField('doc_id'): doc_id, sa_helpers.KeyField.STUDENT: student_id},
+        sa_helpers.KeyStateType.INTERNAL
+    )
```
pmitros, line 51: This should be broken up into functions. This is mixing:

- Low-level data access
- High-level heuristics

This should basically be two function calls to well-named functions, possibly with a dictionary lookup for the latter.

learning_observer/learning_observer/doc_processor.py
```diff
@@ -0,0 +1,235 @@
...
+
+async def check_for_doc_fetch_failure(doc_id):
+    '''Check if we should retry fetching a previously
+    failed document fetch
+    '''
+    now = get_seconds_since_epoch()
+    if doc_id in failed_fetch:
+        last_try = failed_fetch[doc_id]
+        if last_try > now - 300: return False
+    return True
```
pmitros, line 70: I don't quite get what you're trying to accomplish or why. Better docstring?

learning_observer/learning_observer/doc_processor.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+RULES = [
```
pmitros, line 73: Rules are usually combined with AND or OR or similar.

learning_observer/learning_observer/doc_processor.py
```diff
@@ -0,0 +1,235 @@
...
+        end_of_specifier = k.find(doc_specifier) + len(doc_specifier)
+        end_of_id = k.find(',', end_of_specifier)
+        doc_id = k[end_of_specifier : end_of_id if end_of_id != -1 else len(k)]
+        if doc_id == 'None': continue
+        doc_ids.add(doc_id)
+    return doc_ids
+
+
+async def process_documents(docs):
+    results = await asyncio.gather(*(process_document(d) for d in docs))
```
pmitros, line 125: I'm not a big fan of the `gather` operations. This would work better as an asynchronous chain:

```
def fetch_docs(doclist)
  for docid in doc_list:
     yield fetch_doc(docid)
```

```
def process_docs(docs)
  for doc in docs:
     yield process_doc(doc)
```

and `def store_docs()`. This way, every doc is available as soon as it's processed, and perhaps things are a little more responsive. This doesn't give the same feel of "last retrieved 5 hours ago", but you potentially see things come in as they're ready. 

Compute can also run in the background for network and vice-versa.

learning_observer/learning_observer/doc_processor.py
```diff
@@ -0,0 +1,235 @@
...
+    roster_specifier = 'learning_observer.google.roster'
+    rosters = [k for k in keys if roster_specifier in k]
+    matching_teachers = []
+    for roster_key in rosters:
+        roster = await KVS[roster_key]
+        if student in roster['students']:
+            matching_teachers.append(roster['teacher_id'])
+
+    # TODO handle multiple teachers
+    teacher = matching_teachers[0]
```
pmitros, line 177: Same one-to-many comment as above.

learning_observer/learning_observer/doc_processor.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+async def _fetch_document_text(doc_id, creds):
+    '''Fetch the document text from the appropriate
+    Google endpoint.
+    '''
+    async def get_session(request):
+        return {learning_observer.constants.USER: {learning_observer.constants.USER_ID: '12345'}}
+
+    aiohttp_session.get_session = get_session
```
pmitros, line 193: A lot of this mock code should be abstracted out into a function which makes the mock runtime.

learning_observer/learning_observer/doc_processor.py
```diff
@@ -0,0 +1,235 @@
...
+        process_document,
+        {sa_helpers.EventField('doc_id'): doc_id, sa_helpers.KeyField.STUDENT: student_id},
+        sa_helpers.KeyStateType.INTERNAL
+    )
+    await KVS.set(key, output)
+
+
+if __name__ == '__main__':
+    import asyncio
+    asyncio.run(start())
```
pmitros, line 235: I might run this as a process when starting `learning_observer`, rather than as a separate `python` command. This adds devops overhead, which is expensive. 

A good way to do this is to check which processes to start from our settings file. By default, we run all of them, but it's possible to break it up at some point if we ever hit our millionth user.

modules/writing_observer/writing_observer/awe_nlp.py
```diff
@@ -120,7 +120,6 @@ def process_text(text, options=None):
     if options is None:
         # Do we want options to be everything initially or nothing?
         options = writing_observer.nlp_indicators.INDICATORS.keys()
-        options = []
```
pmitros, line 123: I don't understand this. We should explain when we talk.

learning_observer/learning_observer/auth/social_sso.py
bradley-erickson: In the background process we use the doc id to find the student, which finds the teacher (via the roster), which finds the teachers credentials, which are used to fetch the text of the document.

learning_observer/learning_observer/auth/utils.py
bradley-erickson: I believe we will also get an auth exception when the document isn't shared which is likely to happen. I'll see about on adding this to the workflow when the auth expires.

learning_observer/learning_observer/doc_processor.py
bradley-erickson: Will update docstring, but the `failed_fetch` is updated whenever a document fails to fetch the text. We don't want to retry constantly since some documents will never be shared and always fail. This limits it to the checking every 5 minutes.

learning_observer/learning_observer/doc_processor.py
bradley-erickson: Right now the check rules function performs an AND across all rules. I spent a little time thinking about combining them a different way, but went with the current way and awaited some feedback.

modules/writing_observer/writing_observer/awe_nlp.py
bradley-erickson: If no options for AWE components were pass in, did we want to run all of them or none of them? We never came to a conclusion on this and the code sat with both coded in, but the none of them option as the path used.

learning_observer/learning_observer/doc_processor.py
bradley-erickson: I disagree. I think using `gather` here is correct. I could be incorrect, but the asynchronous chain will block the next document from being started. Whereas with the gather operation, while one document is waiting for a response from the Google API or LanguageTool (or another await item) another document can begin its own processing.

Also we ought to create an async wrapper for the NLP AWE components. Currently they are blocking.

learning_observer/learning_observer/doc_processor.py
bradley-erickson: Collin mentioned that currently there are 2 servers running. 1 for incoming events and 1 for viewing the dashboards. He mentioned that the incoming events were causing the rest of the system to slow down a lot.

learning_observer/learning_observer/auth/utils.py
pmitros: I would expect different auth exceptions (even in the text if not the type), but it's worth double-checking. 

We don't want to keep making requests when auth has expired, and we do want to keep track of access permissions.

learning_observer/learning_observer/doc_processor.py
pmitros: Random first thoughts: 

- We should leave a bookmark or todo here. 
- Trying to do unauthorized things every five minutes seems like a really good way to get ourselves flagged and banned by some buggy, automated Google system, so I'm a bit worried about leaving as-is. 
- At the very least, I'd do an exponential backoff, but I'd be inclined to just not retry if it's a sharing issue until we are tracking sharing events. 
- In practice, in the classroom, 100% was shared, so this might be a special case. We can ask the other teachers how this works there.

learning_observer/learning_observer/doc_processor.py
pmitros: My feedback is:
1. This should be documented
2. We should consider a format such that we can extended it later

I'd much prefer `AND(rule, rule)` to `[rule, rule]` or similar but with better syntax.

It's fine to leave this as a bookmark / todo for now, but I'm a little hesitant to not at least do that, since otherwise, we might get other rulesets and be stuck. 

Even if we decide on lists being AND, and having an explicit OR, that should come after some thought (which we don't have time for right now)

learning_observer/learning_observer/doc_processor.py
pmitros: OT: I'd be curious where the `redis` is. Network latency can be high with a non-local db.

modules/writing_observer/writing_observer/awe_nlp.py
pmitros: Ah. Thank you. Makes sense. We should probably do all of them all the time in the current workflow.

learning_observer/learning_observer/doc_processor.py
pmitros: We're both wrong, but we should discuss tomorrow. 

https://docs.python.org/3/library/asyncio-task.html#waiting-primitives

Design requirements: 

1) The thread pool of LanguageTool / AWE / etc. is finite. Running 1000 things at once is bad. Limiting to just one, as I had, is bad too. This should be based on number of cores or similar, so specified in config.

2) I expect we'll be NLP-limited (redis writes will be fast, and hopefully so will Google API calls). If this is the case, we'd want to pull things before shortly before NLP capacity opens up, and store things right after

3) We do want things to be available to teachers once ready, and not with a big delay.

learning_observer/learning_observer/auth/social_sso.py
pmitros: This code makes me unhappy and feels wrong, but the alternatives I've come up with are similarly bad, for other reasons. 

I think the best option I've come up -- not in time for the next study -- is to have a document maintain:

- a list of users associated with it (students and teachers)
- some metadata about those users (e.g. when they last accessed it, permissions issues, etc.)

Might be good to have a conversation at some point. 

We also will need some design documentation here:

- Stepping through all documents feels crazy, as does the path document -> roster -> student
- It's less crazy once you dig into it. Key points are that only _active_ documents are processed, which will be non-obvious very quickly. 

Again, not for next study.

learning_observer/learning_observer/auth/utils.py
bradley-erickson: if we get an auth error, stop trying that document.

learning_observer/learning_observer/doc_processor.py
bradley-erickson: break into a _get_doc_metadata function

bradley-erickson: get doc info, get meta info, check if it needs to be processed (below)

learning_observer/learning_observer/doc_processor.py
bradley-erickson: make it so if the doc is in the failed to fetch dict, return false instead of trying to do the retry every 5 minutes

modules/writing_observer/writing_observer/module.py
```diff
@@ -41,6 +43,21 @@
 
 source_selector = q.call('source_selector')
 
+pmss.register_field(
+    name='get_awe_lt_from_sep_process',
```
pmitros, line 47: Choice and not boolean:

`NLP_Source: choice[compute_in_dag, async_in_process, async_out_of_process]`

Optional class

.awe {
  NLP_Source
}

.language_tool {
  NLP_Source
}

modules/writing_observer/writing_observer/module.py
```diff
@@ -41,6 +43,21 @@
...
+    description='Data can be queried either by system defined execution DAGs '\
+                '(directed acyclic graphs) or user created execution DAGs. '\
+                'This is useful for developing new system queries, but should not '\
+                'be used in production.',
+    default=False
+)
+
+separate_processing_script = learning_observer.settings.module_setting('writing_observer', setting='get_awe_lt_from_sep_process')
+print('* Separate processing script', separate_processing_script)
+nlp_source = 'nlp_sep_proc' if separate_processing_script else 'nlp'
```
pmitros, line 58: Note: The above eliminates code like this, and there is one canonical way of representing things

modules/writing_observer/writing_observer/module.py
```diff
@@ -41,6 +43,21 @@
...
+lt_group_source = 'overall_lt_sep_proc' if separate_processing_script else 'overall_lt'
+lt_single_source = 'single_lt_sep_proc' if separate_processing_script else 'single_student_lt'
```
pmitros, line 60: ... and these become classes. 

.single_student .language-tool {
  NLP_Source
}

And we don't need any logic like this. Again, we want a single representation and source of truth.

modules/writing_observer/writing_observer/module.py
```diff
@@ -63,10 +81,12 @@
...
+        'single_lt_combined': q.join(LEFT=q.variable(lt_single_source), LEFT_ON='provenance.provenance.STUDENT.value.user_id', RIGHT=q.variable('roster'), RIGHT_ON='user_id'),
+        'single_lt_sep_proc': q.select(q.keys('writing_observer.lt_process', STUDENTS=q.parameter("student_id", required=True), STUDENTS_path='user_id', RESOURCES=q.variable("single_student_latest_doc"), RESOURCES_path='doc_id'), fields='All'),
         # overall language tool nodes
         'overall_lt': languagetool(texts=q.variable('docs')),
-        'lt_combined': q.join(LEFT=q.variable('overall_lt'), LEFT_ON='provenance.provenance.STUDENT.value.user_id', RIGHT=q.variable('roster'), RIGHT_ON='user_id'),
+        'overall_lt_sep_proc': q.select(q.keys('writing_observer.lt_process', STUDENTS=q.variable('roster'), STUDENTS_path='user_id', RESOURCES=q.variable("doc_ids"), RESOURCES_path='doc_id'), fields='All'),
+        'lt_combined': q.join(LEFT=q.variable(lt_group_source), LEFT_ON='provenance.provenance.STUDENT.value.user_id', RIGHT=q.variable('roster'), RIGHT_ON='user_id'),
```
pmitros, line 89: This is horrible, but a fine hack for the school pilots this month. We do need to figure out, on a 10,000 foot level, how we handle settings changing query DACs. That's going to take time.

modules/writing_observer/writing_observer/writing_analysis.py
```diff
@@ -139,6 +139,21 @@ async def event_count(event, internal_state):
...
+    return False, False
+
+@kvs_pipeline(scope=gdoc_scope, null_state={})
+async def lt_process(event, internal_state):
+    '''
+    An example of a per-document pipeline
+    '''
+    return False, False
+
+
```
pmitros, line 156: Note: These are a hack. The reducers needs these to be able to query data.

PR #141, opened 2024-05-17
WiP added job loop for running execution dag queries

bradley-erickson:


learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -232,12 +232,22 @@ def exception_catcher(*args, **kwargs):
...
+    Returns a generator of awaitable tasks as they are ready
+    '''
+    tasks = []
+    for value in values:
+        nested_val = get_nested_dict_value(value, value_path)
+        async def _update_overall(func, func_val, val):
+            result = await func(func_val)
+            return _annotate_map_result(result, func, val, value_path)
+        tasks.append(asyncio.create_task(_update_overall(func, nested_val, value), name=nested_val))
+    return asyncio.as_completed(tasks)
```
pmitros, line 250: I have no idea what this code does:

```
        nested_val = get_nested_dict_value(value, value_path)
        async def _update_overall(func, func_val, val):
            result = await func(func_val)
            return _annotate_map_result(result, func, val, value_path)
        tasks.append(asyncio.create_task(_update_overall(func, nested_val, value), name=nested_val))
```

And the overall function, from the name, sounds as if it were just asyncio.as_completed. 

It feels like we'd want something like:

`asyncio.as_completed([meaningful_function_name(stuff) for stuff in somewhere])`

Or similar.

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -278,6 +288,30 @@ def map_serial(func, values, value_path):
         outputs.append(output)
     return outputs
 
+def _annotate_map_result(res, func, value, value_path):
```
pmitros, line 291: res => result

learning_observer/learning_observer/communication_protocol/executor.py
```diff
@@ -680,6 +718,11 @@ async def visit(node_name):
         visited.add(node_name)
         return nodes[node_name]
 
+    # TODO discuss and do something with before merging. How should `clean_json`
+    # process a generator? Using str(generator) makes sense for `clean_json`;
+    # however, we still need to drive the generator. Instead, we could call
+    # `clean_json` after driving the generator and getting the results.
+    return {e: await visit(e) for e in target_nodes}
```
pmitros, line 725: Generically:

1) A generator should be stepped through and converted into a list

2) I don't know if this should happen inside clean_json, as many generators can only be stepped through once

How is this used? I will give two more options:

Client receives:

`[result, result, <awaiting>, result, <awaiting>`

Client receives

`[result, result, old_result, result, old_result]`

Client receives:

`result 5 => {new result 5}

learning_observer/learning_observer/dashboard.py
```diff
@@ -556,6 +560,195 @@ async def websocket_dashboard_handler(request):
...
+
+    query_parameters = None
+    previous_parameters = None
+
+    async def _check_for_diff_params():
+        '''Continually check for our current parameters to change
+        '''
+        nonlocal previous_parameters
+        while previous_parameters == query_parameters:
+            await asyncio.sleep(4)
```
pmitros, line 624: I hate the hardcoded sleeps.

bradley-erickson:
This code is no longer relevant. Much of this has been re-implemented better in #162

PR #142, opened 2024-05-28
added back_to login functionality

bradley-erickson:
When users refresh and are logged out, they will be redirected back to their prior page.

### DIFF
--- a/learning_observer/learning_observer/static/webapp.js
+++ b/learning_observer/learning_observer/static/webapp.js
@@ -104,10 +104,10 @@ requirejs(
 	    d3.select(".lo-google-auth").classed("is-hidden", !config['google_oauth']);
 	    d3.select(".lo-http-auth").classed("is-hidden", !config['http_basic_auth']);
 	    d3.select(".lo-password-auth").classed("is-hidden", !config['password_auth']);
-		d3.selectAll('a').attr('href', function () {
+		d3.select('.lo-google-auth p a').attr('href', function () {
 			const currentHref = d3.select(this).attr('href');
 			if (!currentHref) { return null; }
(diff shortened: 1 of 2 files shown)


learning_observer/learning_observer/static/webapp.js
```diff
@@ -104,6 +104,11 @@ requirejs(
 	    d3.select(".lo-google-auth").classed("is-hidden", !config['google_oauth']);
 	    d3.select(".lo-http-auth").classed("is-hidden", !config['http_basic_auth']);
 	    d3.select(".lo-password-auth").classed("is-hidden", !config['password_auth']);
+		d3.selectAll('a').attr('href', function () {
+			const currentHref = d3.select(this).attr('href');
+			if (!currentHref) { return null; }
+			return currentHref.includes('/auth/login/') ? currentHref + window.location.search + window.location.hash : currentHref;
```
pmitros, line 110: This code is brittle, and will break if we e.g. make server-side changes to URL structure.

learning_observer/learning_observer/static/webapp.js
```diff
@@ -104,6 +104,11 @@ requirejs(
 	    d3.select(".lo-google-auth").classed("is-hidden", !config['google_oauth']);
 	    d3.select(".lo-http-auth").classed("is-hidden", !config['http_basic_auth']);
 	    d3.select(".lo-password-auth").classed("is-hidden", !config['password_auth']);
+		d3.selectAll('a').attr('href', function () {
```
pmitros, line 107: This code is brittle, and will rewrite all anchors. We may very well add other anchors in the future, and we will have a counter-intuitive failure. We always want to select by class or by ID.

PR #147, opened 2024-06-18
WiP: install steps for workshop

bradley-erickson:


learning_observer/util/stream_writing.py
```diff
@@ -185,7 +195,7 @@ async def stream_document(text, ici, user, doc_id):
                     for char, index in zip(text, range(len(text))):
                         command = insert(index + 1, char, doc_id)
                         await web_socket.send_str(json.dumps(command))
-                        await asyncio.sleep(float(ici))
+                        await asyncio.sleep(random.gauss(mu=float(ici), sigma=1))
```
pmitros, line 198: The sigma=1 is not good. 1 second means that this will often be negative, which with `sleep`, clips to zero. It will also often be quite large (a typical ICI is a fraction of a second). So the ICI will be larger than expected. 

I think something like a Reyleigh, Chi, Poisson, or similar distribution might make more sense here.

PR #152, opened 2024-07-16
WiP: New development (based on workshop, coglabs, etc.)

pmitros:


modules/lo_event/lo_event/lo_assess/components/actions.jsx
```diff
@@ -0,0 +1,19 @@
...
+  return Action;
+}
+
+/*
+export function LLMPrompt({children}) {
+  return <></>;
+}
+
+export const reset=() => reduxLogger.setState({});
+*/
```
bradley-erickson, line 19: Is this old code or will something be done with this later?

modules/lo_event/lo_event/lo_assess/components/buildlib.js
```diff
@@ -0,0 +1,145 @@
...
+import glob from 'glob';
+import fs from 'fs';
+import path from 'path';
+import { parseString } from 'xml2js';
+import Mustache from 'mustache';
+import { fileURLToPath } from 'url';
+
+import { config } from './buildConfig.js';
+
+let header = "/* AUTOMATICALLY GENERATED by compileComponents.js: Do not edit. */\n\n";
```
bradley-erickson, line 15: Where is the `compileComponents.js` file? Should this be renamed to `buildlib.js`?

modules/toy-assess/package.json
```diff
@@ -16,6 +16,7 @@
     "bufferutil": "^4.0.8",
     "indexeddb-js": "^0.0.14",
     "jasmine": "^5.1.0",
+    "lo_event": "file:../lo_event/lo_event-0.0.2.tgz",
```
bradley-erickson, line 19: We don't want to link to a specific `.tgz` file.

modules/toy-assess/lo_event.sh
```diff
@@ -0,0 +1,6 @@
+rm -Rf .next/cache/
+pushd ../lo_event/
+npm run prebuild
+npm pack
+popd
+npm install ../lo_event/lo_event-0.0.2.tgz  --no-save
```
bradley-erickson, line 6: Where is this file? How do we build it? Should those commands be included here?

modules/toy-assess/lo_event.sh
bradley-erickson, line 1: What is this file doing?

bradley-erickson:
What is the appropriate way to test this code? What documentation can I follow for doing so? The rough understanding after reviewing the code is:

1. Package `lo_event`
2. Install it in the `toy-assess` context
3. Run the `toy-assess` code

I'm not sure the exact steps to take for any of these items, though not hard to figure out.

modules/lo_event/lo_event/lo_assess/components/buildlib.js
pmitros: It was merged / refactored into `lo_cli.js` and `buildlib.js`. The header should be fixed.

modules/toy-assess/package.json
pmitros: How else would you handle this? Notice that this is a relative link.

modules/toy-assess/lo_event.sh
pmitros: The commands are in the lines above. We go into `../lo_event` and run `npm pack`.

modules/toy-assess/lo_event.sh
pmitros: `lo_event` is the Learning Observer event library (and also `lo_assess` within it is the Learning Observer activity library, which should be factored out eventually). 

* It's a library, and there's no practical stand-alone way to do more extensive development without some use-cases. This is a good set of use cases.
* As we develop those, we often make changes to `lo_event` and `lo_assess`. Quite a lot, actually. 

This script packages `lo_event` into a node package, wipes the `next.js` cache (which often contains relics before changes), and installs it. 

Without this, development of `lo_event` is painful. Even with this, in an ideal case, we would rerun this whenever there were changes to `lo_event` automatically with some kind of watch daemon.

pmitros:
@bradley-erickson  

Instructions:
https://github.com/ETS-Next-Gen/writing_observer/blob/berickson/workshop/docs/workshop.md#better-data-sources-lo_event-and-lo_assess

There are faster ways to do this, with the shell script, and the `npm` commands there now, but that's a good baseline description of how to run this.

modules/toy-assess/package.json
bradley-erickson: No specific method in mind. Using a file for this feels wrong in the long term, but I think is fine given the stage we are at.

modules/toy-assess/lo_event.sh
bradley-erickson: Not familiar with `pushd` and `popd`. I assumed it was building something related to `toy-assess` as opposed `lo_event`. It's clear now that I know what all the commands do, but on first glance it was not.

modules/toy-assess/lo_event.sh
bradley-erickson: Would be beneficial to leave as a comment.

bradley-erickson:
I will work on testing the code some to ensure it works on my machine as well.

bradley-erickson:
Notes about testing the code:
1. Dependency issue when installing LO Event. Specifically around redux, react-redux, redux-thunk. Doing a `npm install --force` fixed this. Not sure how important the versioning is here.
2. Only `npm run build` does not work due to a plethora of linting errors. I previously used this command when building the prototype workflow for serving Next JS applications through Learning Observer.

bradley-erickson:
This pull request is good to merge outside of any additional XML related items @pmitros wants to add.

PR #155, opened 2024-07-24
started adding nextjs related workflows

bradley-erickson:
This allows for serving nextjs applications from LO. To first build the application, follow the steps in the [Toy SBA PR](https://github.com/ETS-Next-Gen/toy-sba/pull/5).

The major issue right now is how we are dealing with paths. The `module.py` variable that describes where to look for NextJS apps should look like:
```python
'''
Next js dashboards
'''
NEXTJS_PAGES = [{
    'path': 'toy_sba/',
}]
```
Where `path` points to the directory within the module that the build NextJS application is in.
In this code, I copied the `out/` directory to the module and renamed it to `toy_sba/`.

### DIFF
--- a/learning_observer/learning_observer/module_loader.py
+++ b/learning_observer/learning_observer/module_loader.py
@@ -54,6 +54,7 @@ AJAX = {}
 
 WSGI = []
 DASH_PAGES = {}
+NEXTJS_PAGES = []
 
 
 def extra_views():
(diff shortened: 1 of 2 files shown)


bradley-erickson:
TODO add documentation

bradley-erickson:
Steps to get this working:
1. Build a nextjs application using `npm run build`
2. Copy the build into a module
3. Define `NEXTJS_PAGES` within the `module.py`
This will build all the non-api portions of the NextJS. The APIs should be re-created or available from the module. There is some work that needs to be about handling API urls within the NextJS application (initially it called `api/llm` but now while in LO it needs to call `module/api/llm`).

PR #159, opened 2024-08-13
Adding logging to document processor service to track updates and fix…

DrLynch:
This PR amends the document processor service to correct one issue with the student_id and to add debug logging to the doc processor so that we can track which documents are actually being handled.


bradley-erickson (approved):
We covered all these changes in our meeting, so they look good to me

Issue #161, opened 2024-08-13
Need Spacy pmss


DrLynch:
Spacy can be directed to require, prefer, or not use GPUs in code.  The settings for this affects how all models and pipeline items get loaded.  In practice it is more efficient to force spacy on this before all other modules are loaded.  At present we use the attached code on our local install:

```
import spacy
#spacy.prefer_gpu()
#debug_log("Preferring GPU Use.")
spacy.require_gpu()
debug_log("Requiring GPU Use.")

```

For robustness this should be behind a pmss feature that allows for 3 settings: require_gpu, prefer_gpu, and no_gpu.  That would allow us to test things separately.

PR #164, opened 2024-09-04
added pmss for spacy gpu preference check

bradley-erickson:
TODO: this still needs testing for the GPU usage. Passing it over to an NCSU student.
My GPU is Cuda version 12.5 and the Spacy version on the system is 3.3. This Spacy version does not support Cuda 12.5.

### DIFF
--- a/modules/writing_observer/writing_observer/awe_nlp.py
+++ b/modules/writing_observer/writing_observer/awe_nlp.py
@@ -4,11 +4,11 @@ This is an interface to AWE_Workbench.
 
 import asyncio
 import enum
-import hashlib
-import time
 import functools
-import os
 import multiprocessing
+import os
+import pmss
+import time
 
 from concurrent.futures import ProcessPoolExecutor
 from learning_observer.log_event import debug_log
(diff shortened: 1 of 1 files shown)


bradley-erickson:
Fixes #161

bradley-erickson:
Waiting on @duckduckdoof to test and make sure it works.

duckduckdoof:
Trying to get a new install of WO working with updated AWE module (with spacy 3.5.4); there is only one dependency conflict after installing AWE modules:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
dash-extensions 1.0.18 requires pydantic<3.0.0,>=2.7.1, but you have pydantic 1.10.18 which is incompatible.
```

Will still be checking if spacy/cuda pmss works.

duckduckdoof:
Had to install WO and point to Arglab versions of components, lexica, languagetool, and spellcorrect.

Finally got WO working with spacy+cupy gpu settings (ran awe_nlp.py main); this works for `process_text()` but not for multithreading functions such as `process_texts_parallel()`.

This issue https://github.com/explosion/spaCy/issues/5507 references that it is a known issue regarding trying to get spacy gpu/cupy stuff working with multithreading.

PR #165, opened 2024-09-06
Scaling doc

pmitros:
@DrLynch @bradley-erickson As per discussion today, I wrote up scaling doc. Please review and then we can merge.

### DIFF
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -1,0 +1,116 @@
+# Scaling Architecture
+
+The goal is for the Learning Observer to be:
+
+* Fully horizontally-scaleable in large-scale settings
+* Simple to run in small-scale settings
+
+It is worth noting that some uses of Learning Observer require
+long-running processes (e.g. NLP), but the vast majority are small,
+simple reducers of the type which would work fine on an 80386
+(e.g. event count, time-on-task, or logging scores / submission).
+
+## Basic use case
+
+In the basic use case, there is a single Learning Observer process
+running. It is either using redis or, if unavailable, disk/memory as a
+storage back-end.
+
+## Horizontally-scalable use-case
+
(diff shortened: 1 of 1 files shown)


bradley-erickson:
Looks good to me.

Brendon-Hablutzel:
comment

Brendon-Hablutzel:
How would the architectures for the horizontally-scalable and task-scalable use cases work together?

According to the doc, for horizontally-scalable, the load balancer would route based on key, but for task-scalable, the load balancer would route based on module/reducer.

Does this imply a set of processes like the following (assuming we split students in half for load balancing purposes and that there are two modules):
- one handling all events from the *first* half of students that go to *module 1*
- one handling all events from the *second* half of students that go to *module 1*
- one handling all events from the *first* half of students that go to *module 2*
- one handling all events from the *second* half of students that go to *module 2*

pmitros:
@Brendon-Hablutzel Consider this use-case:

We have _Learning Observer_ deployed for millions of kids. We have a dozen modules which are well-tested, performant, and which we developed ourselves. We might have a large pool of servers for those modules. Those shard by key within the pool (typically, student). 

An ed-tech wants us to host a half-dozen modules -- stable, but not as fully vetted -- and we deploy their own server pool of perhaps four servers. This maintains isolation from the above pool. 

A researcher develops a module for use in one study in one course in one university. We spin up a tiny instance for that. 

For clarity: These are all _Learning Observer_ servers; we're not shooting for a microservices architecture. 

For more clarity: All this is not something we need to or want to support right  now (unless it becomes someone's pet project). We don't have contributed modules or hundreds of schools. However, on a high-level, we want to directionally move towards where we have server pools specific to organizations (groups of modules), and then within that, to keys (typically students, but we might have e.g. a per-assignment reducer or per-school reducer as well).

PR #166, opened 2024-09-09
Discussion PR: Full codebase versus empty

pmitros:
Experimental PR to be able to discuss the whole code base.


learning_observer/learning_observer/adapters/README.md
pmitros, line 1: We should probably rename this from `adapters` to something which clearly indicates that it is adapting the communication protocol for backwards-compatibility and compatibility with specific standards. Candidate names:

* `protocol_adapters`
* `protocol_compatibility_adapters`
* `event_protocol_compatibility_adapters`
* ... and so in, with increasing length. 

We should distinguish from the communication protocol and other protocols.

PR #178, opened 2024-09-26
Abstracted current gpt responders to their own module

bradley-erickson:


bradley-erickson:
@pmitros this should be ready to merge barring any major changes you wish to see before that. Other items can be added as TODOs

pmitros:
I'm okay with merging this although I'm not a huge fan of this codebase. 

Specifically:

- I generally prefer code less object oriented and more functional
- It's a mistake to have `await lo_gpt.gpt.gpt_responder.chat_completion(gpt_prompt, system_prompt)`. How this is handled may be context-dependent. And `learning_observer.settings.settings['modules']['writing_observer'].get('gpt_responders', {})` isn't the PMSS-friendly way to do things.

If we're making this generic, what I'd really want is, for example:

```.wo .group_a {
  gpt_responder: gpt-4o;
  temperature: 0.5;
}```

```.dynamic-assessment {
  gpt_responder: ollama;
  ollama_server: http://my-little-server.learning-observer.org/
}```

```
[school=ncsu] {
   open_ai_creds: [NCSU-billed account]
}
```

And similar. If this is step zero, of simply putting bricks in the right place, and making a module, it makes sense to me (but it should be documented as scaffolding code). If we're trying to define an API, it makes little sense. 

Test cases would also be helpful, at some point. We're reaching that stage. 

I would also really want a dummy model which can run without devops. It's fine if it just says "I don't know" or calls into [Eliza](https://github.com/wadetb/eliza) or similar.

PR #162, opened 2024-08-21
WIP Highlight Dashboard rewrite

bradley-erickson:
started rewrite of highlight dashboard, dashboard is in a good spot, but need to work on the date flow into it (via the communication protocol)

PR #180, opened 2024-09-27
added dashboard to demo LOConnection and helper utilities

bradley-erickson:


bradley-erickson:
- [ ] Needs documentation

bradley-erickson:
- [ ] Add ability to manually call open and close on the websocket

bradley-erickson:
[ ] Data manager component needs testing now that #162 has been merged in

modules/toy-assess/src/app/websocket/page.js
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  const [userData, setUserData] = useState({});
+  const [errors, setErrors] = useState({});
+
+  const { readyState, message, error, sendMessage, openConnection, closeConnection } = LOConnection({ url: 'ws://localhost:8888/wsapi/communication_protocol', dataScope });
+  const handleDataUpdate = ({ dataObject, errors }) => {
+    setUserData(dataObject);
+    setErrors(errors);
+  };
```
pmitros, line 30: ```userData = useLOConnection([...]);```

`use` is a special thing in React.

pmitros: https://react.dev/learn/reusing-logic-with-custom-hooks#extracting-your-own-custom-hook-from-a-component

pmitros: redux would also be a reasonable way to handle state here, as we're moving the codebase in that direction.

pmitros: https://github.com/ETS-Next-Gen/writing_observer/blob/master/modules/lo_event/lo_event/lo_assess/selectors.js <-- This would be a reasonable way to go. Just a thin wrapper around `useSelector`

modules/toy-assess/src/app/websocket/page.js
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <div>
+      <h1>WebSocket Connection Page</h1>
+      <div>
+        <LOConnectionLastUpdated message={message} readyState={readyState} />
+        <LOConnectionDataManager message={message} onDataUpdate={handleDataUpdate} />
+      </div>
+      <div>
+        <Button onClick={() => sendMessage(JSON.stringify(dataScope))} disabled={readyState !== WebSocket.OPEN}>
```
pmitros, line 39: * I'm wondering if we should be exposing this is a websocket to a dashboard writer. 
* LOConnection returns a lot of state. `readyState` `message` and `error` should be in a hook, so it can signal that re-renders trigger when these change. If this is implicit, it will be confusing. This is an error in the code (even if it works). 

Read about hooks and redux state manangement. Consider what to move into redux and how we want redux. Etc. 

This might be LOConnection updating redux, and something like:

```
lo = LOConnection(...);
error = useLOConnectionError(lo);
state = useLOConnectionState(lo);
```

Or similar.

modules/toy-assess/src/app/websocket/page.js
```diff
@@ -0,0 +1,64 @@
...
+      execution_dag: 'writing_observer',
+      target_exports: ['docs_with_nlp_annotations'],
+      kwargs: decoded
+    }
+  };
+
+  const [userData, setUserData] = useState({});
+  const [errors, setErrors] = useState({});
+
+  const { readyState, message, error, sendMessage, openConnection, closeConnection } = LOConnection({ url: 'ws://localhost:8888/wsapi/communication_protocol', dataScope });
```
pmitros, line 26: If this were a React component, this would be `React.createElement(LOConnection);`. This is probably being treated as a function call. That works, but it masks the state management inside, which will confuse users of this,.

pmitros: I believe that this would fail:

```
if(running) {
   conn = LOConnection();
}
```

Hooks have to be called the same number of times and in the same order in each component.

modules/toy-assess/next.config.js
```diff
@@ -11,20 +11,26 @@ const nextConfig = {
...
         options: {
           // Add your babel presets here (e.g., @babel/preset-react)
-          presets: ['@babel/preset-react'], // Assuming you use React
-        },
-      },
-    })
-    return config
+          presets: ['@babel/preset-react'] // Assuming you use React
+        }
+      }
```
pmitros, line 24: In config files, the commas are now standard JavaScript style. 

dict = {
   a: b,
   c: d,
   e: d,
};

This is so:
* when we add a new line, the diff only shows the new line
* re-ordering lines, commenting out lines, etc. doesn't break things

pmitros:
Where we might be headed:

```
import { someStuff };

export function myComponent(data) {
  // When we need to add some interactivity, it might be a line like:
  const dropdown = useLOConnection(query);
  // ...
  return (<>{data}</>);
};

// Syntax below would be different.
//
// Perhaps we'd want a standard way to define display modes
myComponent.displayModes = [latestGrid([some stuff]), // All students current work, for the whole class
                            assignmentSingle()];      // Select a student and assignment from a dropdown
```

bradley-erickson:
TODOs before merging

- [ ] Test to confirm the hooks work
- [ ] Move `Websocket` states to their own variable in `useLOConnection`
- [ ] Move `rendertime` function to util file
- [ ] Assess naming of exported `useLOConnection` items in `useLOConnectionDataManager`. I feel like they may want to be exported as a named item, `ws.sendMessage` instead of just `sendMessage` to help keep the naming separate.

modules/lo_event/lo_event/lo_assess/components/utilities/useLOConnectionDataManager.jsx
```diff
@@ -0,0 +1,134 @@
...
+        console.error('Failed to parse incoming message:', e);
+      }
+    }
+  }, [message]);
+
+  return {
+    connection,
+    data: state.data,
+    errors: state.errors,
+  };
```
bradley-erickson, line 133: @pmitros review this.

We want to expose the endpoints on the `useLOConnection` hook used within this data manager hook. What do you think is the proper way to expose these?

Initially I had `...connection` which exposed each item (`sendMessage, message, error, connectionStatus, openConnection, closeConnection`) as their own thing. I didn't like how the naming could possibly clash with the data manage. I switched to `connection` so they are all namespaced when being used. Maybe I'm thinking too hard about this. See the `page.js` example for current usage.

modules/lo_event/lo_event/lo_assess/components/utilities/useLOConnectionDataManager.jsx
```diff
@@ -0,0 +1,134 @@
...
+  data: {},
+  errors: {},
+};
+
+export const useLOConnectionDataManager = ({ url, dataScope }) => {
+  const { message, ...connection } = useLOConnection({ url, dataScope });
+  const [state, dispatch] = useReducer(dataReducer, initialState);
+
+  useEffect(() => {
+    if (message) {
```
pmitros, line 111: I am very confused by this in useEffect. 

Shouldn't we be passing a callback into useLOConnection or something?

modules/lo_event/lo_event/lo_assess/components/utilities/useLOConnection.jsx
```diff
@@ -0,0 +1,95 @@
...
+      closeConnection();
+    };
+  }, [url]);  // Include `url` as a dependency in case it changes and requires a reconnection
+
+  // Function to send a message via WebSocket
+  const sendMessage = (message) => {
+    if (clientRef.current && connectionStatus === LO_CONNECTION_STATUS.OPEN) {
+      clientRef.current.send(message);
+    } else {
+      console.warn('WebSocket is not open. Ready state:', connectionStatus);
```
pmitros, line 90: This will lead to bugs and timing issues. I think it will be pretty common to see code where sendMessage is called while the connection is opening. Queues are pretty typical here. Not the rock-hard industrial queue in lo_event, but just a JS list.

PR #181, opened 2024-10-03
WiP Remaining conversions to support PMSS

bradley-erickson:


bradley-erickson:
Last item to get converted is feature flags. This one is a little interesting since sometimes its a simple true/false and sometimes we want the specific sub-settings for that feature flag.
Both `auth_headers_page` and `merkle` expect a dictionary to be returned.

bradley-erickson:
Per a conversation with Piotr. Perhaps we try implementing something similar to CSS Keyframes (used in animations to define the css at different points along the animation) to support lists or dictionaries.
So maybe something like:
```css
@list xyz {
  0 { setting: 123; }
  1 { setting: 456; }
  ...
}
```
OR 
@dict xyz {
  item1 { setting: 123; }
  item2 { setting: 456; }
  ...
}
```

PR #192, opened 2024-10-29
Addressed weak reference for created tasks in asyncio

bradley-erickson:
While reading up on some other Asyncio items, I stumbled across this piece of [documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

> Important
> 
> Save a reference to the result of this function, to avoid a task disappearing mid-execution. > The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

```python
background_tasks = set()

for i in range(10):
    task = asyncio.create_task(some_coro(param=i))

    # Add task to the set. This creates a strong reference.
    background_tasks.add(task)

    # To prevent keeping references to finished tasks forever,
    # make each task remove its own reference from the set after
    # completion:
    task.add_done_callback(background_tasks.discard)
```

This pull request implements this for the Websocket Dashboard API.

### DIFF
--- a/learning_observer/learning_observer/dashboard.py
+++ b/learning_observer/learning_observer/dashboard.py
@@ -588,6 +588,7 @@ async def websocket_dashboard_handler(request):
     previous_client_query = None
     batch = []
     lock = asyncio.Lock()
+    background_tasks = set()
 
     async def _send_update(update):
         '''Send an update to our batch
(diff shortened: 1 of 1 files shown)


pmitros:
I'm okay merging this, since it's a clear improvement, but it's far from perfect. When tasks finish, ideally, they'd be removed from the set() so it doesn't keep growing forever. 

Ideally, though, this shouldn't be necessary. Normally, references are kept since the outer code which then acts on it. I suspect you're doing something like:

```
async def handle_item():
   do_computation()
   update_result()

await do_all_of(handle_item, items)
```

Where it might be:

```
def handle_item():
  return do_computation()

async for return_value in do_all_of(handle_item, items):
   update_result(return_value)
```

It's more complex than the pseudocode above, obviously (especially since we don't know whether `do_computation` or `update_result` is faster, so we might want to update several results).

bradley-erickson:
They are currently removed from the set after completion. That's what the `task.add_done_callback(background_tasks.discard)` does. `add_done_callback` calls the passed in function with the future object, in this case `task`, as the only argument.

pmitros:
Okay. Let's merge without a TODO. 

If we get back to the code later, perhaps we can rework the logic if we remember.

PR #196, opened 2024-11-11
WiP: further classroom fixes

bradley-erickson:
Creating this PR so we can more easily see the diffs being made.


modules/lo_event/lo_event/lo_event.js
```diff
@@ -49,13 +46,30 @@ async function initializeLoggers () {
...
-    initialized = INIT_LOGGERS_READY;
+    initialized = INIT_STATES.LOGGERS_READY;
   } catch (error) {
-    initialized = INIT_ERROR;
+    initialized = INIT_STATES.ERROR;
     debug.error('Error resolving logger initializers:', error);
   }
 }
 
+export async function compileMetadata(metadataTasks) {
```
bradley-erickson, line 56: Docstring would be nice

modules/lo_event/lo_event/lo_event.js
```diff
@@ -98,35 +112,43 @@ export function init (
...
-    () => setFieldSet([{ source, version }]));
+  initialized = INIT_STATES.IN_PROGRESS;
+  currentState = currentState
+    .then(initializeLoggers)
+    .then(() => setFieldSet([{ source, version }]))
+    .then(() => compileMetadata(metadata));
+  if(sendBrowserInfo) {
+    // In the future, some or all of this might be sent on every
+    // reconnect
+    logEvent("BROWSER_INFO", getBrowserInfo());
```
bradley-erickson, line 145: `"BROWSER_INFO"` would be better as a variable.
Currently in `lo_event` all the verbs are variables. In the extension, they are just listed as strings. They should probably all be defined as a variable somewhere, so we can reference them in both javascript and python code.

modules/lo_event/lo_event/metadata/storage.js
```diff
@@ -0,0 +1,59 @@
+export function getStorageMetadata(storage, keys = null) {
```
bradley-erickson, line 1: Docstring please

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -3,10 +3,35 @@ import * as disabler from './disabler.js';
...
 
-export function websocketLogger (server) {
+function wsHost(overrides = {}, loc=window.location) {
+  const { hostname, port, path, url } = overrides;
+  const loServer = localStorage.getItem('lo_server');
+  if (loServer) {
+    console.log("Overriding server from localStorage");
+    return loServer;
+  }
+  const protocol = loc.protocol === 'https:' ? 'wss://' : 'ws://';
```
pmitros, line 13: This should be wss -> wss to. This should never come up since it's grounding in window.location, but still, we should be resilient.

modules/lo_event/lo_event/disabler.js
```diff
@@ -106,6 +106,14 @@ export async function retry () {
   }
   action = DEFAULTS.action;
   expiration = DEFAULTS.expiration;
-  storage.set({ [DISABLER_STORE]: { action, expiration } });
+  // HACK when expiration is set to null, this gets called constantly
```
bradley-erickson, line 109: @pmitros please review this change

modules/lo_event/lo_event/util.js
```diff
@@ -492,3 +389,31 @@ export function formatTime(seconds) {
...
+ * This function dispatches an event in the appropriate context for
+ * our environment.
+ *
+ * When working in an extension, we want to send a message via the
+ * `chrome.runtime` object.
+ *
+ * When working in a browser, we want to dispatch the event via the
+ * `window` object.
+ */
+export function dispatchCustomEvent(eventName, detail) {
```
bradley-erickson, line 403: This is how we can dispatch events in both the normal context, using `window`, and the extension context, using `chrome.runtime`.

pmitros: Thank you. This is nice. I suspect it would make sense to switch this around, though, and have:

  ```if (typeof window !== "undefined") {
      // Web page: dispatch directly on window
      window.dispatchEvent(event);
 } if (typeof chrome !== "undefined" && chrome.runtime && chrome.runtime.sendMessage) {
      // Chrome extension background script: use chrome.runtime to send messages
      chrome.runtime.sendMessage({ eventName, detail }, (response) => {
        if (chrome.runtime.lastError) {
          console.warn(`No listeners found for event, ${eventName}, in this context.`);
        }
      });
  } else {
      console.warn("Event dispatching is not supported in this environment.");
  }
}```

The extension is the corner case. The default case should check for a normal browser environment and use that if available.

modules/lo_event/lo_event/disabler.js
```diff
@@ -106,6 +115,15 @@ export async function retry () {
...
+  // NOTE: If the client continuously sends messages to the server and
+  // keeps generating new messages, we may hit a rate limit depending on
+  // the `storage` API in use. To avoid this, we only call the `set` method
+  // when the new value differs from the existing value in `storage`.
+  storage.get([DISABLER_STORE], (result) => {
+    const currentValue = result[DISABLER_STORE];
+    if (!currentValue || currentValue.action !== action || currentValue.expiration !== expiration) {
+      storage.set({ [DISABLER_STORE]: { action, expiration } });
+    }
+  });
```
bradley-erickson, line 127: We need to reset the storage for future initializations to know if they are allowed to transmit events or if they need to wait.
Is using the `.get` and comparing what is in there the appropriate way to avoid constantly pinging the `.set` api?

modules/lo_event/lo_event/websocketLogger.js
pmitros: Oh, sorry. This should be something along the lines of:

```
protocol_map = {wss: 'wss', ws: 'ws', http: 'ws', https: 'wss'}`
protocol = protocol_map[loc.protocol]
```

To make a long story short, there's a lot which should be done to make this function more robust in the future, one of which being able to take in `loc` as a string. 

The logic below is wonky too:

```
  // This logic might be moved into wsHost, so it's a little bit more
  // consistent.
  if(!server) {
    server = wsHost();
  } else if(typeof server === 'object') {
    server = wsHost(server);
  }
  // else the server is likely already a url string
```

Ergo, the comment. As it is, if it's already a string, overrides don't override, which is counterintuitive. 

That, in turn, changes the definition of webSocketLogger to something like `export function websocketLogger (location, overrides)` or similar. 

I didn't have the energy to pipe all of that through the system.

extension/writing-process/src/background.js
```diff
@@ -205,15 +219,12 @@ async function reinjectContentScripts() {
...
-        loEvent.logEvent('chrome_identity', { chrome_identity: result });
-        loEvent.logEvent('metadata_finished', {})
-    }
-});
+// NOTE the Learning Observer server expects a `chrome_identity` event.
+// This event should include the auth information for a given user.
+// The auth information is included in our metadata via the `chromeAuth`
+// metadata task.
+// There may be cleaner ways to handle this.
+loEvent.logEvent('chrome_identity', {});
```
bradley-erickson, line 227: This is what's needed to ensure the auth workflow operates properly. The server expects a `chrome_identity` event which used to come from the `profileInfoWrapper` (which now exists within `chromeAuth`).

modules/lo_event/lo_event/websocketLogger.js
```diff
@@ -70,36 +104,8 @@ export function websocketLogger (server) {
...
-           * is updated and we have a new handshake protocol.
-          */
-          queue.enqueue(JSON.stringify(metadata));
-          queue.enqueue(JSON.stringify({ event: 'chrome_identity', chrome_identity: result }));
-        }
-      });
-    } else {
-      firstConnection = false;
+    if(Object.keys(metadata).length > 0) {
+      queue.enqueue(JSON.stringify(metadata));
```
bradley-erickson, line 108: This still needs to be tested. I think we may need to include a `chrome_identity` event to get re-authed.

modules/lo_event/lo_event/lo_event.js
```diff
@@ -49,13 +46,36 @@ async function initializeLoggers () {
...
+/**
+ * Executes and compiles metadata tasks into a single metadata object.
+ *
+ * When initializing `lo_event`, clients can set which metadata items
+ * they wish to include.
+ */
+export async function compileMetadata(metadataTasks) {
+  const taskPromises = metadataTasks.map(async task => {
+    try {
+      const result = await Promise.resolve(task.func());
```
bradley-erickson, line 65: I cleaned up this line. I'm not sure that we need to include the `async` key in the metadataTasks anymore. The code used to read (which was also backwards)

`const result = await (task.async ? task.func() : Promise.resolve(task.func()));`

extension/writing-process/src/background.js
pmitros: This is where we should make a parallel change in the protocol in `incoming_student_event.py`. The PR is unfinished here. It should initially support both methods, and then deprecate the old method. 

I will mention: There is a large number of related methods there, which are essentially "Show a token" which might come from a storage, Chrome, or other places. Those could be made uniform.

modules/lo_event/lo_event/lo_event.js
pmitros: Huh. I did use the old code, which was clearly broken. I'm curious why it worked the old, broken way.

pmitros:
@bradley-erickson I have a million comments, which isn't really practical for a code review, mostly reflecting the in-progress stage of this code. I slept on it, and my thoughts:

* Bringing this up to high code quality is a lot of work, and so shouldn't be a blocker on a merge
* We will specifically need to work through the logic on a lot of the more complex pieces in more depth (websocketlogger queuing, disabler, etc.)
* This will be easier with the existing cleanups

I think we should squash and merge, but continue development from where we are. 

I think the major question are protocol changes. You may have tested for this already, but the new way of collecting metadata might have intentionally or inadvertently changed the protocol. If so, we'd receive event logs in a new format. 

What would be really helpful would be to think through how we might test this. We have a lot of auth protocols, and it'd be nice to have end-to-end tests that they all work between `lo_event` and `pmss`. However, this isn't quite a blocker on a merge, since the only event auth we have serious data from is the extension and nginx.

PR #201, opened 2024-11-27
Fixes to reconstruction code

PDDeane:
It turns out we were missing two critical commands 'rvrt' and 'rplc', both of which were used to replace large chunks of text, often the entire document. 'rplc' sometimes appears as the first command in a document, in which case it reflects the creation of a document from a template that the author didn't write. To handle this I wrote a replace() function to dispatch from 'rvrt' and 'rplc', plus null actions dispatched for a bunch of suggestion commands.

In the course of getting this code ready to put a pull request in for, I seem to have created some whitespace differences I can't track down. They shouldn't matter, but sorry!

PR #202, opened 2024-12-02
Pdeane reconstruct pr

bradley-erickson:
New PR for #201 so I can more easily make changes.
 
 > It turns out we were missing two critical commands 'rvrt' and 'rplc', both of which were used to replace large chunks of text, often the entire document. 'rplc' sometimes appears as the first command in a document, in which case it reflects the creation of a document from a template that the author didn't write. To handle this I wrote a replace() function to dispatch from 'rvrt' and 'rplc', plus null actions dispatched for a bunch of suggestion commands.

> In the course of getting this code ready to put a pull request in for, I seem to have created some whitespace differences I can't track down. They shouldn't matter, but sorry!

- Paul

### DIFF
--- a/modules/writing_observer/writing_observer/reconstruct_doc.py
+++ b/modules/writing_observer/writing_observer/reconstruct_doc.py
@@ -327,7 +327,7 @@ dispatch = {
                      #what text was NOT written by the author,
                      #logging the text buffer after the initial
                      #rplc action will give you that.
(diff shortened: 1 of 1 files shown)


pmitros:
I had some comments, but they were addressed in Brad's commit. Looks good to merge. 

Thank you so much for this code. This will make things a lot better.

PR #203, opened 2024-12-03
WiP General module updates

bradley-erickson:
This PR is a work in progress for adding various module improvements to the system, mainly versioning and testing.


Makefile
```diff
@@ -1,16 +1,31 @@
...
 	cd learning_observer && python learning_observer
 
-venv:
-	# This is unnecessary since LO installs requirements on install.
-	# pip install --no-cache-dir -r requirements.txt
-
-# install commands
-install: venv
+# Install commands
+install:
```
bradley-erickson, line 28: We should rename this to be more explicit.

Makefile
```diff
@@ -1,16 +1,31 @@
+# TODO rename these packages to something else
 PACKAGES ?= wo,awe
 
+help:
+	@echo "Available commands:"
+	@echo ""
+	@echo "run                  Run the learning_observer Python application."
+	@echo "install              Install the learning_observer package in development mode."
```
bradley-erickson, line 8: Include some mention about the environment which it is being ran in.

Makefile
```diff
@@ -87,8 +99,24 @@ linting-node:
...
 	# Finished linting
 
+# Build commands
 build-writing-ext:
 	# Installing LO Event
 	cd modules/lo_event && npm install & npm link lo_event
 	# Building extension
 	cd extension/writing-process && npm install && npm run build
+
+build-dist:
```
bradley-erickson, line 109: More explicit about which package we are building.

Makefile
```diff
@@ -87,8 +99,24 @@ linting-node:
 linting: linting-setup linting-python linting-node
 	# Finished linting
 
+# Build commands
 build-writing-ext:
```
bradley-erickson, line 103: `build-wo-chrome-extension` or similar

Makefile
```diff
@@ -87,8 +99,24 @@ linting-node:
...
 	cd extension/writing-process && npm install && npm run build
+
+build-dist:
+	# Building distribution for package
+	pip install build
+	# Switching to package directory
+	cd $(PKG) && python -m build
+
+# TODO we may want to have a separate command for uploading to testpypi
+upload-to-pypi: build-dist
```
bradley-erickson, line 116: Add echo commands about running these manually and testing them.

docs/testing.md
```diff
@@ -2,6 +2,37 @@
 
 Testing is an essential part of software development to ensure the reliability and stability of your code. In the Learning Observer project, we currently support Python testing, using the pytest framework. This document provides an overview of how the testing framework is used in the project.
 
+## Preparing a Module for Testing
+
+Each module should define a `test.sh` at the root of the module's directory. This script should define how to run the tests for the module.
```
pmitros, line 7: This is a horrible hack. We should have explicit, recommended ways to run tests,. Document, don't fix.

pmitros: We will want a finite, well-defined set of tests:

* Client / dashboard tests
  * Basic: we're not getting 500 errors on any pages (which can be semi-automatic from routes)
  * Advanced: things work as expected.
*  Python unit tests
* Server-side integration tests
* Data pipeline tests
  * The very explicit models, like reducers, are designed to support this
  * Also, query language, etc.
  * This is related to replicability, which is important in science and in debugging in ways we discussed

This is a big pile of work, but for each of these, we should have an opinionated, recommended way to do things, including in the example template package. That gives flexibility in how we run tests.

pmitros: This won't happen overnight, or even in the near future, but likely when we get some sort of open science grant or similar. But we want to document this as a code direction, move that way, and make sure we don't build up too much cruft around ./test.sh.

bradley-erickson:
I have one issue with the current implementation of versioning. It is very difficult to have one package state that it needs a specific version of another (`foo` requires `bar==1.1.0`).

pmitros:
Suggestion:

```
learning_observer/util ⟶ /scripts
/scripts/hooks
```

bradley-erickson:
- [ ] Update learning observer to use a pyproject.toml

PR #204, opened 2024-12-10
extension fixes to extract document history

bradley-erickson:


pmitros:
This looks good to me, except for the lack of error handling. If there is no docs history, the exception should be sent to the server in place of the docs history. This most likely means we either return a token with the docs history or an error token. 

Another is to raise an exception and to propagate it up to where we request the Google Docs history. 

But in either case, we should send something to lo_event, either with the history, or with as much information as we have about why it's not there for future debugging.

This is important since this code is brittle. We are parsing Google JS which might change. It's also helpful to document why this is brittle and what we're doing. 

In either case, with those changes, it looks good for a merge. Without those changes -- if there is time pressure -- it still makes sense to merge since the existing code is broken.

extension/writing-process/src/writing.js
```diff
@@ -142,29 +142,31 @@ function is_string(myVar) {
...
+                if (match && match[1]) {
+                    const data = JSON.parse(match[1]); // Parse the JSON string
+                    return data.info_params.token;
+                }
+            } catch (error) {
+                console.error('Error parsing _docs_flag_initialData:', error);
+            }
+        }
     }
+    return null;
```
pmitros, line 169: Just below, there is a very obsolete doc string which is confusing.

PR #208, opened 2025-01-02
updated requirements

bradley-erickson:


pmitros:
Looks good to merge

A lot of the requirements should be removed, and it is helpful to have documented what they were for.

PR #209, opened 2025-01-07
WiP: Save state in LO Event

bradley-erickson:
This pull request implements the ability to save and load state within LO Event. This includes both clientside and server-side storage.

Branch: Pbrost/coglabs deployment


learning_observer/learning_observer/incoming_student_event.py
```diff
@@ -447,6 +448,35 @@ async def filter_blacklist_events(events):
...
+                source = event['source']
+                activity = event['activity']
+
+            # Save, fetch, or ignore (continue)
+            if event['event'] == 'save_blob':
+                await learning_observer.blob_storage.save_blob(
+                    user_id, source, activity,
+                    event['blob']
+                )
+            elif event['event'] == 'fetch_blob':
```
pmitros, line 474: @bradley-erickson  I am wondering how this will play with client-side code. Ideally, this would be spent back before we're done with metadata / init / headers / auth (before the events start streaming). 

We should chat in the morning about whether the client should request this when opening the websocket or whether this should be automatically sent.

learning_observer/learning_observer/blob_storage.py
```diff
@@ -0,0 +1,36 @@
...
+        sa_helpers.KeyStateType.INTERNAL
+    )
+    return key
+
+async def fetch_blob(user_id, source, activity):
+    '''Fetch the blob from the KVS
+    '''
+    key = _make_key(user_id, source, activity)
+    kvs = learning_observer.kvs.KVS()
+    return await kvs[key]
```
bradley-erickson, line 29: What happens if the key is missing from the kvs? Should probably return that there was nothing (empty object or `None`).

pmitros: That's fine for now. `null` and not `None`. 

```
$ json.dumps(None)
'null'
```

In the long term, we should support having a default value. In the long term, I would envision this being part of a module rather than the core. But we've got a long way before we're ready for that. 

(Also, note that's different from `json.dumps("null")` which is `'"null"'`)

modules/lo_event/lo_event/reduxLogger.js
```diff
@@ -120,19 +172,30 @@ function set_state_reducer(state = {}, action) {
   return action.payload;
 }
 
+function set_setting_reducer(state = initialState, action) {
```
bradley-erickson, line 175: @pbrost could you refresh me on what this code is used for?

PR #223, opened 2025-05-09
202505/berickson/lti integration

bradley-erickson:
LTI integration there are still some TODOs in code


learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+    '''
+    # TODO read the provider configs from settings somehow
+    provider_configs = []
+    for config in provider_configs:
+        provider_name = config['name']
+        PROVIDERS[provider_name] = config
+
+
+# ======================
+# JWKS Utilities
```
pmitros, line 55: This comment should explain what JWKS stands for, and what it does.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+def get_public_key(provider: str, kid: str):
+    '''Get RSA public key from JWKS'''
+    jwks = JWKS_CACHE.get(provider, {})
+    if jwk := find_jwk(jwks, kid):
+        return jwt.algorithms.RSAAlgorithm.from_jwk(jwk)
+    return None
+
+
+# ======================
+# JWT Validation
```
pmitros, line 88: Spell out and explain JWT.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+        provider_name = config['name']
+        PROVIDERS[provider_name] = config
+
+
+# ======================
+# JWKS Utilities
+# ======================
+
+
+async def fetch_jwks(config: dict) -> dict:
```
pmitros, line 59: We should be deliberate and explicit about whether we adopt LLM coding style (e.g. type hints, comment format, etc.). 

Right now, this code just feels LLM-generated and out-of-step with our main codebase. We can decide to adopt the same conventions elsewhere, but it should be a stated decision.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+        'client_id': 'YOUR_CLIENT_ID',
+        'redirect_uri': 'https://yourdomain.com/lti/launch/canvas',
+        'private_key_path': './keys/canvas/private.pem'
+    }
+    '''
+    # TODO read the provider configs from settings somehow
+    provider_configs = []
+    for config in provider_configs:
+        provider_name = config['name']
+        PROVIDERS[provider_name] = config
```
pmitros, line 51: This code seems to have a lot of wrapper functions which don't do much. Why do we want to copy things over from config? Shouldn't we just be calling into config? 

This kind of thin wrapper means:

a) People familiar with our main codebase will have an internal config system to learn

b) If we ever go dynamic (e.g. config comes from a database), this code will break.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+# ======================
+# JWKS Utilities
+# ======================
+
+
+async def fetch_jwks(config: dict) -> dict:
+    '''Fetch JWKS from provider's endpoint'''
+    async with aiohttp.ClientSession() as session:
+        async with session.get(config['jwks_url']) as response:
+            return await response.json()
```
pmitros, line 63: We should evaluate, on all of these, whether:

a) We want a wrapper or just  aiohttp.ClientSession().get(config['jwks_url']).json() in-line (with the proper awaits as necessary)
b) The wrapper should be global, or inside the function where this is used

It feels like this code has a lot of layers and abstractions which might not be adding much value.

This, I would definitely place inside of get_jwks, since it's not even clear the difference between get and fetch. It could be a function inside, or it could be just 1-3 lines of code.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+async def get_jwks(provider: str) -> dict:
+    '''Get cached JWKS or fetch fresh if not available'''
+    if provider not in JWKS_CACHE:
+        config = get_provider_config(provider)
+        JWKS_CACHE[provider] = await fetch_jwks(config)
+    return JWKS_CACHE[provider]
+
+
+def find_jwk(jwks: dict, kid: str) -> dict:
+    '''Find JWK by key ID'''
```
pmitros, line 75: More context. Find where? For what purpose? 

I'd probably inline this, and also be mindful of handling duplicates (not just missing).

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+        public_key,
+        algorithms=['RS256'],
+        audience=claims.get('aud'),
+        issuer=claims.get('iss')
+    )
+
+
+# ======================
+# Client Assertion
+# ======================
```
pmitros, line 119: This needs about a paragraph. I have no idea, on a first read, what "Client Assertion" means.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+def read_private_key(path: str) -> str:
+    '''Read PEM-formatted private key from file'''
+    with open(path, 'r') as f:
+        return f.read()
```
pmitros, line 125: Inline as `open(path).read()`

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+
+def create_assertion_payload(client_id: str, token_url: str) -> dict:
+    '''Create JWT payload for client assertion'''
+    now = int(time())
+    return {
+        'iss': client_id,
+        'sub': client_id,
+        'aud': token_url,
+        'iat': now,
+        'exp': now + 300,
```
pmitros, line 136: Comments. What is iss? What is sub? How do the two differ? 300, I assume, is 5 minutes? What's jti? 

I'd also inline or place inside of generate_assertion. This code has a lot of cross-references which are hard to follow.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+
+def generate_assertion(config: dict) -> str:
+    '''Generate signed client assertion JWT'''
+    private_key = read_private_key(config['private_key_path'])
+    payload = create_assertion_payload(config['client_id'], config['token_url'])
+    return jwt.encode(payload, private_key, algorithm='RS256')
+
+
+# ======================
+# OIDC Handlers
```
pmitros, line 149: Paragraph. Assume reader has never heard of OIDC.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+    '''Create user session data from LTI claims'''
+    roles = claims.get('https://purl.imsglobal.org/spec/lti/claim/roles', [])
+    is_instructor = 'http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor' in roles
+    return {
+        constants.USER_ID: claims['sub'],
+        'email': claims.get('email'),
+        'name': claims.get('given_name'),
+        'family_name': claims.get('family_name', ''),
+        'picture': claims.get('picture', ''),
+        'role': learning_observer.auth.ROLES.TEACHER if is_instructor else learning_observer.auth.ROLES.STUDENT,
```
pmitros, line 212: Roles are usually a list, since a user might have several.

learning_observer/learning_observer/auth/lti_sso.py
```diff
@@ -0,0 +1,297 @@
...
+        'authorized': is_instructor
```
pmitros, line 213: Authorized for what?

learning_observer/learning_observer/auth/lti_sso.py
pmitros: They're also contextual. I might be a student in one class, and an instructor in another.

learning_observer/learning_observer/canvas.py
```diff
@@ -0,0 +1,129 @@
...
+    # Create the API routes and get back the functions
+    global_functions = initialize_and_register_routes(
+        app=app,
+        endpoints=ENDPOINTS,
+        api_name='canvas',
+        feature_flag_name='canvas_routes'
+    )
+
+    # Add the functions to the module's global namespace
+    globals().update(global_functions)
```
pmitros, line 31: This kind of magic injection is usually very bad form. It breaks a lot of things, including things like code analysis and optimization tools. 

If we have this, we should document why. From this point in the code, it's not even clear what is being injected into globals()! 

It's also bug-prone, since we're making some dependency on order-of-operations. Reviewing, it's not clear when 
`initialize_and_register_canvas_routes` is called, and when injected functions will get called. But this is the kind of place it's easy to have a bug down-the-line.

learning_observer/learning_observer/canvas.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def _extract_course_id_from_url(url):
```
pmitros, line 34: This should be it's own function (as it is), but moved inside of clean_course_list, instead of module-global.

learning_observer/learning_observer/external_apis.py
```diff
@@ -0,0 +1,293 @@
...
+            aiohttp.web.get(
+                e._local_url(),
+                make_ajax_raw_handler(e.remote_url)
+            )
+        ])
+
+    return result_functions
+
+
+def register_cleaner_factory(endpoints):
```
pmitros, line 261: This comment does not explain what a "cleaner" is, or why this exists.

learning_observer/learning_observer/canvas.py
pmitros: So as a rule-of-thumb, magic should be generally avoided. 

We do have magic in APIs we expect to be used many times, to simplify those APIs for users. That's a fine place to use a lot of introspection, decorators, self-modifying code, etc. since we're complicating our lives to simplify that of our users. 

In the case of our own code, and especially relatively bounded code, we should try to keep code understandable.

learning_observer/learning_observer/external_apis.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+async def raw_api_ajax(runtime, target_url, key_translator=None, cache=None, cache_key_prefix=None, **kwargs):
+    '''
+    Make an AJAX call to an API, managing auth + auth.
+
+    * runtime is a Runtime class containing request information.
+    * target_url is typically grabbed from ENDPOINTS
+    * key_translator is a dictionary to translate API keys to internal keys
+    * cache is an optional cache object
```
pmitros, line 77: Explain what the caching is for. Performance? Off-line debugging?

learning_observer/learning_observer/external_apis.py
```diff
@@ -0,0 +1,293 @@
...
+    return [f[1] for f in string.Formatter().parse(format_string) if f[1] is not None]
+
+
+async def raw_api_ajax(runtime, target_url, key_translator=None, cache=None, cache_key_prefix=None, **kwargs):
+    '''
+    Make an AJAX call to an API, managing auth + auth.
+
+    * runtime is a Runtime class containing request information.
+    * target_url is typically grabbed from ENDPOINTS
+    * key_translator is a dictionary to translate API keys to internal keys
```
pmitros, line 76: An example of the overall call input / output would be helpful. 

Even looking at the code, this: `return learning_observer.util.translate_json_keys(response_data, key_translator)` is less than intuitive. There are a lot of levels of indirection here.

learning_observer/learning_observer/canvas.py
```diff
@@ -0,0 +1,129 @@
...
+        return match.group(1)
+    return None
+
+
+@register_cleaner("course_list", "courses")
+def clean_course_list(canvas_json):
+    '''
+    The LTI integration Canvas uses for auth occurs on a
+    course by course level. This cleaner wraps the current
+    course in a list.
```
pmitros, line 45: On all of these, it'd be helpful to explain input / output in the comment.

learning_observer/learning_observer/google.py
```diff
@@ -1,49 +1,32 @@
...
+import learning_observer.kvs
+import learning_observer.util
 
+from learning_observer.external_apis import Endpoint, initialize_and_register_routes, register_cleaner_factory
 
+# Cache for Google API responses
 cache = None
 
-
+# Google field name mapping
```
pmitros, line 29: It's confusing to have a list be called a mapping

pmitros:
I reviewed the code. I think I'm in a good position to have a discussion about it tomorrow. A lot of the comments were my initial impressions, and so based on an incomplete understanding (and may not be correct, but rather point to places where we need better documentation / variable names). 

Where I see some friction: 

* In the original code, the whole setup with the cleaners was designed to make local functions which call into AJAX APIs. We were still figuring out which APIs we would call, and it was nice to be able to call a broad number of them and add / remove them as needed. This is especially true across things like Google Docs, where we might want all sorts of oddball calls. 
* In this code, the AJAX code is designed to give a few generic calls across systems. E.g. I can grab a roster from Google or from Canvas. I'm not sure we'll be doing things like setting up a Canvas assignment or whatnot. If not, I'm not sure this kind of magic is warranted here. But perhaps it is; we'd want to understand more about how these might be called downstream. 

Other issues:

* Major security issue is embedding things in error messages we don't want going to the user. 
* A lot of one-off functions should either not be in functions, or more likely, be inside of the function they are used in. Rather than:
'''
def _foo():
   ...
def bar():
   _foo()
'''

We should have:
'''
def bar():
   def foo():
      ...
   foo()
'''

So we know it's scoped locally. There are a lot of levels of indirection in the code
* Docstrings should describe where things fit in the overall structure (and not repeat the function name)
* Function names should be better

In general, it often feels like there are too many levels of indirection. 

The code looks mostly mergeable, though, once we fix the issue of error messages.

learning_observer/learning_observer/rosters.py
```diff
@@ -86,15 +87,16 @@
...
 pmss.register_field(
     name='source',
     type='roster_source',
+    # TODO update this source information
     description='Source to use for student class rosters. This can be\n'\
                 '`all`: aggregate all available students into a single class\n'\
                 '`test`: use sample course and student files\n'\
                 '`filesystem`: read rosters defined on filesystem\n'\
-                '`google_api`: fetch from Google API',
+                '`google|schoology|...`: fetch from specific API',
```
bradley-erickson, line 99: Clean this up

learning_observer/learning_observer/rosters.py
```diff
@@ -349,22 +351,26 @@ def init():
...
+    #     raise learning_observer.prestartup.StartupCheck(
+    #         "Settings file needs a `roster_data` element with a `source` element. No `source` element found."
+    #     )
+    if roster_source in ['test', 'filesystem']:
         ajax = synthetic_ajax
+    # TODO both canvas and google don't use the ajax function when the query is made
     elif roster_source in ["google_api"]:
         ajax = google_ajax
+    elif roster_source in ["canvas_api", 'schoology_api']:
+        pass
```
bradley-erickson, line 369: Does keeping Ajax and None do anything? We should maybe set it to an empty function instead.

learning_observer/learning_observer/integrations/__init__.py
```diff
@@ -0,0 +1,17 @@
+import learning_observer.integrations.canvas
+import learning_observer.integrations.google
+import learning_observer.integrations.schoology
+
+INTEGRATIONS = {}
+
+
+def register_integrations(app):
+    # TODO go through settings to determine which of these we want to enable on the system
```
bradley-erickson, line 9: Resolve this TODO.

Perhaps a setting like register_schoology_endpoints (boolean)
One for canvas as well that's a list of items we want

learning_observer/learning_observer/routes.py
```diff
@@ -246,6 +246,26 @@ def register_auth_webapp_views(app):
                 handler=learning_observer.auth.social_handler),
         ])
 
+    # TODO check settings for LTI providers
+    # If they are available, then register these routes. Skip otherwise.
+    # TODO build provider syntax based on available providers
+    if True:
```
bradley-erickson, line 252: Resolve this TODO

learning_observer/learning_observer/settings.py
```diff
@@ -225,7 +225,8 @@ def initialized():
 # Not all of these are guaranteed to work on every branch of the codebase.
 AVAILABLE_FEATURE_FLAGS = [
     'uvloop', 'watchdog', 'auth_headers_page', 'merkle', 'save_google_ajax', 'use_google_ajax',
-    'google_routes', 'save_clean_ajax', 'use_clean_ajax'
+    'google_routes', 'save_clean_ajax', 'use_clean_ajax',
+    'canvas_routes', 'schoology_routes'
```
bradley-erickson, line 229: Leave comment discussing how we should handle updating this list in the future. Always needing to update this list feels like it could be improved upon.

learning_observer/learning_observer/static/webapp.js
```diff
@@ -115,6 +115,7 @@ requirejs(
 	function load_login_page() {
 	    d3.select(".main-page").html(mustache.render(login, config['theme']));
 	    d3.select(".lo-google-auth").classed("is-hidden", !config['google_oauth']);
+	    d3.select(".lo-lti-auth").classed("is-hidden", !config['lti_auth']);
```
bradley-erickson, line 118: Add the ability for the lti button to refresh the page.

PR #235, opened 2025-09-15
WiP: updated extension to only connect when relevant tabs are present

bradley-erickson:
Initial testing seems okay, but needs more rigorous testing before we merge.

Ready for an initial code review.

### DIFF
--- a/extension/writing-process/src/background.js
+++ b/extension/writing-process/src/background.js
@@ -9,7 +9,9 @@ var RAW_DEBUG = false;
 /* This variable must be manually updated to specify the server that
  * the data will be sent to.  
 */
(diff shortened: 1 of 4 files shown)


extension/writing-process/src/background.js
```diff
@@ -28,34 +28,54 @@ import { localStorageInfo, sessionStorageInfo } from 'lo_event/lo_event/metadata
...
+    );
+    loEvent.go();
+    loEventActive = true;
+    loEvent.logEvent('extension_loaded', {});
+    logFromServiceWorker('Extension loaded');
+}
+
+function stopLogger () {
+    if (!loEventActive) return;
+    loEvent.logEvent('terminate', {});
```
bradley-erickson, line 70: We should handle this type of event on the server - i.e. stop the connection / close files / ...

